PR #63, opened 2018-01-15
Export gltf animations into a single GLTFAnimation

noalak:
Also remove some .cache files


tkazik:
very nice...works for me!

PR #67, opened 2018-01-18
Correcting some project file referencing mistakes 

Selmar:
- Correcting some project file referencing mistakes.
- Set platform target to x64 for max2016 as well.


deltakosh:
Pinging @noalak for validation

noalak:
It's ok, good catches Selmar

PR #88, opened 2018-02-04
Added x64 platform again and did some cleanup.

Selmar:
**NOTE** It all compiles, but is untested, as I have no access to 3ds max '17 or '18. Can't see a reason why these changes would break anything, but it seems reasonable to verify.

Some project cleanup:
- Added x64 platform in addition to AnyCPU.
- Default platform is now x64 again. AnyCPU only needs to be there to compile custom forms so they can be viewed and edited.
- Fixed release builds (allow unsafe compilation).
- Cleanup unused hasExporter variable.
- Removed some unsafe code and void ptrs that seemed useless.


deltakosh:
Did you try to edit the forms in VS?

Selmar:
Yes, it works as expected. They don't show or update in VS unless you (re)compile them for x86.

deltakosh:
Fine

Issue #76, opened 2018-01-25
Morpher + Multi/Sub-Object Material don't work together


Tinderfox:
Steps:
1. Make a box (box1)
2. Convert to poly
3. Duplicate it (box2)
4. Morpher modifier on box1
5. Edit box2 and set a morph animation (optionally export and run after to see the previewer preview the animation now)
6. Put a material on box1
7. Put a different material on one face of box1
8. Export and run now, to see that animation doesn't play

I think that means there's an issue with the exporter. Importing this model into Godot 3 RC3 crashes it. This doesn't happen if either the morpher modifier is removed or box1's material is changed to a non- multi material. Having both on the same model causes the babylon.js preview in "Export & Run" not to show the animation, which I take to mean something's breaking in the exportation, as trying to import this same model into Godot 3 RC3 crashes it. This happens whether you export to .gltf or .glb

Issue #49, opened 2017-12-13
Meshes exported to gltf seem to always be flagged wtih BLENDMODE instead of opaque


deltakosh:

deltakosh:
I think this one was done right?

PR #92, opened 2018-02-06
Fix issues and update dlls

noalak:
Fix issue #76 
Fix issue #49 
Update dlls including last @Selmar updates


Selmar:
Just curious, what's the rationale behind setting OPAQUE explicitly instead of leaving it out?

deltakosh:
I wanted it to make sure all engines will have it explicitely and because some artists were complaining about not finding it in the file

Selmar:
I doubt any engine would "forget" to set a default to opaque, but I can see why an artist would be worried. 👍

deltakosh:
And artists are respectful people I don't want to mess with :)

Issue #51, opened 2017-12-20
Support hair & fur export


kutomer:
Hello :)
I think it will be very useful if we will be able to export [3dsmax hair & fur modifer](https://knowledge.autodesk.com/support/3ds-max/learn-explore/caas/CloudHelp/cloudhelp/2016/ENU/3DSMax/files/GUID-4BCFC5E4-266F-43AA-A552-77CC241F9D1E-htm.html) as [Babylon.FurMaterial](http://doc.babylonjs.com/extensions/fur)

in the meantime do we have any other way to export fur & hair from 3dsmax to Babylon.js?

Thanks!

deltakosh:
This is a good idea. So far the only way would be to do it at load time in JS :(

kutomer:
Thanks a lot for the quick reply!
Do you think that it's something can be added to a backlog or something?
Would love to help to develop this feature but guidance is needed.

deltakosh:
Obviously :)
You can perhaps familiarize yourself with the C# code of the exporter. Then we can discuss where in the json file we can export it (it is not a big deal as we can export all kind of materials)

kutomer:
Do you have any documentation for the exporter which I can begin with?

deltakosh:
I would say tags just loadibg the VS project should be a good start
Project is pretty straightforward

kutomer:
Thanks,

1. I couldn't find a way to debug the plugin (read that I need to be a certified Autodesk developer), the way I'm doing it now is with `RaiseMessage` and for that, I need to rebuild and relaunch 3dsmax what makes the development process much slower.
2. I guess I'm doing something wrong but `meshMode.IGameObject.NumModifiers.ToString()` returns 0 for me, have any idea?

noalak:
Hi kutomer,

1) Doing the same
2) Hair and Fur modifier is a [World Space Modifier](http://help.autodesk.com/view/3DSMAX/2017/ENU/?guid=__files_GUID_DA7227A8_576A_4D3D_9254_4DFA1C49B08E_htm). Don't really know much about it but [find attached the code](https://github.com/BabylonJS/Exporters/files/1611117/hairAndFur_code.txt) you are probably looking for.

With that you get all parameters from Hair and Fur modifier. Seems like there is no interface provided by the API, so you need to retreive each parameter by id. You need to adapt the getter method (GetFloat, GetColor...) to the parameter type of each parameter, otherwise it would return the default value (0 for a float).

[...]

PR #96, opened 2018-02-09
Support Hair and Fur 3dsmax modifier

kutomer:
#51


noalak:
Thoughts:
1) Good stuff!

2) Verions 2015 & 2018 raise an error because you encountered a technical issue for those versions or you just ignored/couldn't test them?

3) I don't really see why a for loop (or 2) would be required. just pick what you need directly:
`var paramBlock = modifier.GetParamBlock(0);
spacing = (int)Math.Round(paramBlock.GetFloat(XXXX, 0, 0) / 3); // Cut length`
In case i'm wrong, in the paramBlock switch, could you replace the name of each parameter by its actual paramId so it is not language sensitive.

4) I'm surprised you don't have a warning raised everytime as i would have expected the tip color to never be null (pale brown by default) line 74. But if it works...Nevertheless good idea for the warning messages keep them even if you remove the for loop!

kutomer:
@noalak 
1. Thanks.

2. 
2015 - doesn't support Hair and Fur modifier at all.
2018 - I couldn't check it, I'm using Mac, our modeler have 3dsmax 2017 installed, so I didn't want to just guess, sorry for that.

3. will do.

4. I also thought that this warning will be raised every time :)

furthermore, I've changed the parameters a little bit so the exported Babylon mesh will be closer to what you see in 3dsmax.
That's a bit confusing but I think that it will be better to use the root thickness to determine the density (and not the density) that's because density in Babylon works in a way that the lower the density there are fewer hairs and each is therefore thicker.

I'm fixing it now and will push it within the next few hours.

Thanks again :)

kutomer:
done @noalak

noalak:
Perfect!
Will test with 2018 and 2017 French versions tomorrow.
Can you share a sample max file ? or add it directly to this PR as you wish.

kutomer:
Thanks :)
attached sample file.
[sample.zip](https://github.com/BabylonJS/Exporters/files/1721016/sample.zip)

noalak:
I've downloaded your project and the sample file.

Couldn't test the end result as the sandbox doesn't have external reference to FurMaterial.

Anyway tested export process:
1) 3ds max 2017: The tip color warning is displayed everytime. Export is fine though. If you find a workaround to keep the warning message but to only print it when necessary it would be really nice. Otherwise, just remove it and it will be added to the documentation.
2) 3ds max 2017 French version: modifier is not recognized. Would be nice to find a solid fix to share your work with wilder range of users at some point. For now you can add an extra condition `|| modifier.Name == "Chevelure et Pelage"`
3) 3ds max 2018: all right!

Quick fix 1) & 2) and good to go for me!

Congratz and thanks for your work :)

deltakosh:
@noalak feel free to add a reference to mat libs in the sandbox

deltakosh:
I've just added reference to mat library in sandbox so you should be able to test :)

kutomer:
@noalak done.
Thanks :)

noalak:
Here are the results for the sample file provided by kutomer.

By default, it's very dark. Any reason for that @kutomer @deltakosh ?
default

With higher light intensity:
withhigherlightintensity

We can see the gap between the fur created from the side and the one created from the top. I've no idea on what we should have, so I let you comment on it @deltakosh 

Babylon file: [cylinder_rose-blue-flower-rose-blooms.zip](https://github.com/BabylonJS/Exporters/files/1726774/cylinder_rose-blue-flower-rose-blooms.zip)

kutomer:
@noalak , I don't see it's too dark. (in my application, I don't have light with high intensity)
Maybe it's the black furColor?

deltakosh:
Adding @luaacro for comments on fur rendering

julien-moreau:
Hey sorry for the delay!
Is it rendered with the property "furColor" equals to new Color3(1, 1, 1); ?

noalak:
You can access fur material properties in the babylon file provided in the zip (beautify it [here](https://codebeautify.org/jsonviewer)).

furColor is (0, 0.4549, 0.6039). Setting it to (1,1,1) make it brighter but still darker than i would expect (need to increase light intensity).

kutomer:
Any news? :)

julien-moreau:
Sorry for the delay, I need to investigate for sure because I have no idea for now :(
Anyway, I focus 100% on the problem tomorrow until I find the solution!

julien-moreau:
Hey just reproduced, it looked like the normals were badly computed in the shader
Fixing it right now :)

julien-moreau:
Pull Request created https://github.com/BabylonJS/Babylon.js/pull/3851

kutomer:
@julien-moreau Thanks a lot :)
@noalak Are we good now?

noalak:
@julien-moreau 
Using the babylon file _cylinder_rose-blue-flower-rose-blooms.zip_ linked previously i don't see any difference light-wise. Is that normal?
The fix applied to normal is about the last sample file given by kutomer (carpet)?

kutomer:
It didn't work for me as well...

kutomer:
@julien-moreau Just checking in, but... any news?

julien-moreau:
Dam! Just saw your friendly reminders!
Checking ASAP, so sorry !

kutomer:
@julien-moreau nothing to be sorry for, your help is very much appreciated :)

deltakosh:
Any news?

julien-moreau:
image
According to this screenshot, black ambient occlusion color is the wanted behavior, to help the eye seeing the "thicker" or "reflief", like "depth

julien-moreau:
If disabled, this will render: 
image

kutomer:
@julien-moreau just want to make sure that I understand - the occlusion color is black (can I control this color somehow?) so when I specify "furColor" (or texture) the actual fur color will be from that shade to black? (but mostly this color)

The problem is that this behavior makes the hair look much darker than expected (in the specified example above a completely white fur looks gray)

If I understood it correctly I think a better solution will be to support something like "hair base color \ texture" (same as 3dsmax Hair & Fur modifier)
* the hair "lower" parts will be in the "base color" and the higher in "tip color" I think that's a better solution for hair perspective.

if the above solution is too complicated I think the second option (disabled occlusion color - completely white) will be more predictable and it will be up to user to create a fur texture that looks realistic. 

What do you guys think?

julien-moreau:
I'm adding the "furOcclusion" property to control the occlusion strength: https://github.com/BabylonJS/Babylon.js/pull/4235

I took a moment to read about fur rendering and it looks like most realistic fur renderers do not use the method we have now. They mainly use a combination of complex shaders including geometry shaders :(

kutomer:
@julien-moreau awesome thanks!
@noalak So... are we good to go?

julien-moreau:
Have you tested to let me know if it fits your request ? :)
Thanks!

kutomer:
@julien-moreau sorry for the late response again - but on which scale does the `furOcclusion`? I've tried few possible numbers (0, 1...) all seem the same to me.
Am I missing something here?

julien-moreau:
Can you go over the interval [0, 1] and test with value 2 for example?

kutomer:
Yeah sure, it's the same model I've uploaded on Feb 27.
With 0,1,2 furOcclusion: 

screen shot 2018-05-13 at 11 09 26
screen shot 2018-05-13 at 11 08 24
screen shot 2018-05-13 at 11 07 43

kutomer:
all look exactly the same to me, the image is a bit bigger that's all.

deltakosh:
So we can merge? :) (we need to fix conflicts)

PR #100, opened 2018-02-23
Export materials

noalak:
Export Phong, PhongE, Blinn and StingrayPBS materials to babylon and gltf formats
Update babylon to gltf conversion from left to right handed system, doing it at object level rather than root level
Fix export when selecting non DAG objects


noalak:
@deltakosh 

An error occurered when pushing my local repo from TortoiseGit:
erreurtortoisegit
This screenshot has been made after pushing a 2nd time in a row. But the error was identical.
Writing "Success" in both cases.

Is it meaningful ?

deltakosh:
I have no idea
can you check that all files are in the PR?

noalak:
All files in project solution are here. Didn't check for samples.

deltakosh:
Ok I've merged. Just try to pull from a fresh folder to make sure :)

PR #101, opened 2018-02-27
Added option to automatically create/update .manifest file

FunFetched:
Enabling this option will tell the exporter to create or update the scene's associated .manifest file automatically, using a version number derived from the system date/time. Adds a checkbox to the scene data display. Defaults to false.


deltakosh:
Pinging @palmer-jc

Palmer-JC:
I am not sure why the entire JSON_exporter.py is deleted and re-added. Also do not know if this is a valid manifest file.  Probably works.  Zip file & version need changing, but not necessarily in this commit.

PR #104, opened 2018-02-28
Update Unity exporter to export metal roughness texture map.

justinctlam:
- Add additional properties related to metallic pbr material
- Remove the need to check for ShadowMap component because Unity meshes have a shadowCastingMode that serve the same purpose
- Add support to export Metallic texture map. B channel contains metallic, G channel contains roughness.

### DIFF
--- a/Unity 5/EditorToolkit/Assets/Babylon/Scripts/ExporterWindow.cs
+++ b/Unity 5/EditorToolkit/Assets/Babylon/Scripts/ExporterWindow.cs
@@ -41,7 +41,7 @@ namespace Unity3D2Babylon
         bool showPreview = false;
         bool showCompiler = false;
         int buildResult = 0;
-        string defaultProjectFolder = String.Empty;
+        public string defaultProjectFolder = String.Empty;
         string guiProjectFolder = String.Empty;
         public bool previewThread = false;
 
--- a/Unity 5/EditorToolkit/Assets/Babylon/Entities/BabylonPBRMaterial.cs
+++ b/Unity 5/EditorToolkit/Assets/Babylon/Entities/BabylonPBRMaterial.cs
@@ -95,6 +95,9 @@ namespace BabylonExport.Entities
         public bool useRoughnessFromMetallicTextureGreen { get; set; }
 
         [DataMember]
+        public bool useMetallnessFromMetallicTextureBlue { get; set; }
+
+        [DataMember]
         public bool useAlphaFromAlbedoTexture { get; set; }
 
         [DataMember]
@@ -133,6 +136,9 @@ namespace BabylonExport.Entities
         [DataMember]
         public int maxSimultaneousLights { get; set; }
 
+        [DataMember]
+        public int sideOrientation { get; set; }
+
         public BabylonPBRMaterial() : base()
         {
             SetCustomType("BABYLON.PBRMaterial");
@@ -153,8 +159,12 @@ namespace BabylonExport.Entities
             // Default Null Metallic Workflow
             metallic = null;
             roughness = null;
-            useRoughnessFromMetallicTextureAlpha = true;
-            useRoughnessFromMetallicTextureGreen = false;
+            useRoughnessFromMetallicTextureAlpha = false;
+            useRoughnessFromMetallicTextureGreen = true;
+
+            useMetallnessFromMetallicTextureBlue = true;
+
+            sideOrientation = 1;
 
             microSurface = 0.9f;
             useMicroSurfaceFromReflectivityMapAplha = false;
--- a/Unity 5/EditorToolkit/Assets/Babylon/Scripts/ExporterMetadata.cs
+++ b/Unity 5/EditorToolkit/Assets/Babylon/Scripts/ExporterMetadata.cs
@@ -1124,8 +1124,9 @@ namespace UnityEditor
 
             this.metallic = null;
             this.roughness = null;
-            this.useRoughnessFromMetallicTextureAlpha = true;
-            this.useRoughnessFromMetallicTextureGreen = false;
+            this.useRoughnessFromMetallicTextureAlpha = false;
+            this.useRoughnessFromMetallicTextureGreen = true;
+            this.useMetallnessFromMetallicTextureBlue = true;
 
             this.microSurface = 0.9f;
             this.useMicroSurfaceFromReflectivityMapAplha = false;
--- a/Unity 5/EditorToolkit/Assets/Babylon/Scripts/SceneBuilder.Lights.cs
+++ b/Unity 5/EditorToolkit/Assets/Babylon/Scripts/SceneBuilder.Lights.cs
@@ -49,22 +49,19 @@ namespace Unity3D2Babylon
                 {
                     if (gameObject.layer != ExporterWindow.PrefabIndex)
                     {
-                        if (!gameObject.IsLightapStatic()) {
-                            var shadowmap = gameObject.GetComponent<ShadowMap>();
-                            if (shadowmap != null && shadowmap.runtimeShadows == BabylonEnabled.Enabled)
+                        if (!gameObject.IsLightapStatic())
+                        {
+                            var renderer = gameObject.GetComponent<Renderer>();
+                            var meshFilter = gameObject.GetComponent<MeshFilter>();
+                            if (meshFilter != null && renderer != null && renderer.shadowCastingMode != ShadowCastingMode.Off)
                             {
-                                var renderer = gameObject.GetComponent<Renderer>();
-                                var meshFilter = gameObject.GetComponent<MeshFilter>();
-                                if (meshFilter != null && renderer != null && renderer.shadowCastingMode != ShadowCastingMode.Off)
-                                {
-                                    renderList.Add(GetID(gameObject));
-                                    continue;
-                                }
-                                var skinnedMesh = gameObject.GetComponent<SkinnedMeshRenderer>();
-                                if (skinnedMesh != null && renderer != null && renderer.shadowCastingMode != ShadowCastingMode.Off)
-                                {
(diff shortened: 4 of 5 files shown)


Unity 5/EditorToolkit/Assets/Babylon/Scripts/ExporterWindow.cs
```diff
@@ -41,7 +41,7 @@ public class ExporterWindow : EditorWindow
         bool showPreview = false;
         bool showCompiler = false;
         int buildResult = 0;
-        string defaultProjectFolder = String.Empty;
+        public string defaultProjectFolder = String.Empty;
```
justinctlam: I'm exposing this because it would be useful for tools that are using this script to set their own output path.

deltakosh:
Pinging @MackeyK24 for additional review

MackeyK24:
Sounds Kool... I have already made this change to the up coming 3.2 Toolkit... Also note, in the OnEnable function we-reset the defaultProjectFolder... Not sure what version you have but if you are reseting do something like this OnEnable:

            if (String.IsNullOrEmpty(this.defaultProjectFolder)) {
                this.defaultProjectFolder = Tools.GetDefaultProjectFolder();
            }


Also the new UI has buttons to set the default project Any whre you want and will save that in the project config

screen shot 2018-02-28 at 2 50 45 pm

justinctlam:
Where can I get access to the latest changes?

MackeyK24:
Coming soon... still got a few animation and particle system issues to work out.

But you can do the 'public' thing and check the OnEnable not reseting the defaultProjectFolder until
the 3.2 is ready

justinctlam:
Any chance I can get a preview so I can provide some feedback? :) I don’t need animation or particle systems, mainly pbrmaterials.

MackeyK24:
I am still kind of working out small little Standard Shader (BABYLON.PBRMaterial) issues as well. I would KILL for someone to actually help me work on some of this stuff. If you know anything about all the little 'properties' on the PBRMaterial and how they effect rendering ... ie useMetallicRoughness from alpha vs not and use 'Metalness' from red vs blue channel

I have a ton of new code where I THINK i got the whole metallic/roughness vs metallic/texture vs specular/glossiness workflows (and how they effected by the skybox environment lighting). How i encode the various options according to properties 'Metalness' from red or blue or roughness from alpha.

I need somebody who really know this stuff on the babylon side of the PBRMaterial but who also can understand WHERE i am pulling the information from and how i am serializing all this into the scene file as a material... WHOOO :)


Anyways, if you wanna work wit me on this hit me up, we can GOTOMETTING and talk about this stuff and close it up (ill send my developer version of toolkit)

Email Me

MackeyK24@gmail.com

deltakosh:
Hey guys, what should we do with this PR?

justinctlam:
Depends... if we think the new exporter is coming soon, then we can discard this PR. If it will take a long time, then this exporter is broken without my PR. I will let you guys call this. Thanks.

MackeyK24:
Justin... I have a few questions about the default usage of the metallic blue and roughness green settings you are using. Are those just the default ones from the sketchlab exporter or have you actually got the RIGHT metallic texture output working...

Why do you (or sketchlab) alway put metallic in blue... Is that a default unity thing or wouldn't that be determined if that texture had what you want to use at metallic encoded in the blue channel... what if it is in the red... Can it be in the red in unity...

Same thing for roughness in the green instead of the alpha... What what this determination... what the texture is encoded with... or does unity always use the green for roughness... 

or is this stuff you actually worked out by trial and error..

Again... what does it look like your mods... can you show me what it is supposed to look like.

You can contact me via email (Although a Phone Call would be great so we can put these issues to bed once and for all ... without all the email back and forth with takes forever)

:)

justinctlam:
Q) Why do you (or sketchlab) alway put metallic in blue
A) I think the default parameters in Babylon.js puts it in blue, but it is also the GLTF specification standard. Since this is unrelated to GLTF, you can put it in blue or leave it in red. I think red is default. This is the same for roughness. Sketchfab exports GLTF and green is the GLTF specification. Again, since we are outputting to Babylon.js, you can change it to be alpha as well.

MackeyK24:
You can cancel this PR. I have the updates in the up coming 3.2 version

MackeyK24:
Hey Justin... FYI . alot of the PBR glossiness issues not quite looking right was because we were BOTH
setting metallic and roughness to 1 when using our metallicTexture (weather using blur or red). That is how the GLTF Exporter worked. But apparently in babylonjs, if you are using a metallaicTexture you MUST set pbr.metallic AND pbr.roughness = NULL and ENCODING everything into the texture image.

So if you want to SCALE the roughness or smoothness you WOULD NOT set ANY value to pbr.roughness but instead APPLY a scale or mutiply (_GlossMapScale) to the ALPHA coming from the source metallicTexture to the GREEN (or alpha) of the resulted exported MetallicTexture

Same thing with Specular Gloss Maps... You need to re-encode the scaled (_Glossiness) alpha back
into the resulted exported SpecularGlossMap

This looks beautiful now... Check these shots out

screen shot 2018-03-30 at 3 36 17 pm

screen shot 2018-03-30 at 3 37 17 pm

screen shot 2018-03-30 at 3 37 49 pm


Check the new DumpReflectivityAndGlossiness functions in the up coming 3.2 release

sebavan:
@MackeyK24  actually, when using a metallic texture roughness and metallic are used as factors multiplied against the value fetch from the texture. 1 equivalent to identity simply means to not scale the values in the texture. 

You are exported data looks great !

MackeyK24:
So a value of 1 mean no scale, what is zero and how high can it be scaled

Also... is it the same with micro surface and the encoded alpha data for
specular gloss maps

sebavan:
0 would force everything to 0 as it is a multiplication between the scalar factor and the map values.

We are not limiting the max as you can see in the code below:

```
// Colors
                    if (defines.METALLICWORKFLOW) {
                        PBRMaterial._scaledReflectivity.r = (this._metallic === undefined || this._metallic === null) ? 1 : this._metallic;
                        PBRMaterial._scaledReflectivity.g = (this._roughness === undefined || this._roughness === null) ? 1 : this._roughness;
                        this._uniformBuffer.updateColor4("vReflectivityColor", PBRMaterial._scaledReflectivity, 0);
                    }
                    else {
                        this._uniformBuffer.updateColor4("vReflectivityColor", this._reflectivityColor, this._microSurface);
                    }
```

MackeyK24:
Sorry Justin... I stand corrected on the metal and roughness scaling

MackeyK24:
Just clarify,  does micro surface get multiply by specular alpha... beacuse
i was saling that in texture as well

sebavan:
Yes they are multiplied together in spec gloss workflow. From the shader code:
```
float microSurface = vReflectivityColor.a;
vec3 surfaceReflectivityColor = vReflectivityColor.rgb;

vec4 surfaceReflectivityColorMap = texture2D(reflectivitySampler, vReflectivityUV + uvOffset);
surfaceReflectivityColor *= toLinearSpace(surfaceReflectivityColorMap.rgb);
surfaceReflectivityColor *= vReflectivityInfos.

[...]

#ifdef MICROSURFACEFROMREFLECTIVITYMAP
    microSurface *= surfaceReflectivityColorMap.a;
    microSurface *= vReflectivityInfos.
#else
[...]
```

MackeyK24:
Yo Sebavan... I dont get this....

How can roughness scale go BEYOND 0 and 2...

if i want to make it scale TOTALLY rough i would think you would simple set
roughness to 2.0 but that does not work...you have to go up past 3 to get
that flat look...

HOW if roughness is 0 to 1 . and when using a metallic texture roughness
should be 1 for NO scale.... THEN HOW THE HECK does a value of 2 NOT set
tot totally rough...

I dont get... i seem to have better looking results when i just set metalic
and roughness to NULL and handle encoding glossiness into the texture....
the only problem is i have no real rhyme or reason  how the roughness scale
is supposed to to work

sebavan:
Because it is a MULTIPLICATION so if the texture is having 0 as a value, the scale will have no impact.

And think about your textures are usually int so smallest none 0 would be 1 / 255 which requires a scale of 255 to reach one also as fetching from the value is usually BI or TRI linear you could even require a lot more.

MackeyK24:
So then really when using metallic texture... roughness is NO LONGER a
normalize 0 to 1 value... But instead you use 0 to 255 ...

where 0 is TOTALLY rough and 255 would be MAX smoothness ... ???

MackeyK24:
or would 0 be MAX shininess and 255 be totally rough for the scaling with
metallic texture ???

sebavan:
In metal roughness pipeline, the roughness value from the material is MULTIPLY by the value in the texture.

The final value after multiplication should be between 0 and 1 and 0.001 * 1000 = 1 so if the value in the texture is really small, the factor could really high with still a normalized output.

MackeyK24:
Shit... now im really lost... I dont see why that value is not just a 0 to
1 ... meaning 0 has nothing 1 is full... i guess that is what is throwing
me off

But the fact remains, i have NO IDEA what i should use for roughness math
then:


Lets stay i want to apply a 25% SCALE UP of the shininess...

var scaleFactor = 0.25f;

pbr.roughness = WHAT IS THE CALCULATION - ???


Sorry bro... im not getting it, please bare with me 🤔

MackeyK24:
That us exactly what u was doing but i thought you said before i should
just encode 1 - the raw alpha value for the texture and use pbr.roughness
to scale... being 1 equal no scale and to go up or down just minus or add
to the 1 for roughness... which you can do but the math us way more
difficult getting that scale to work rather than just encode into the
texture... which us what i an going back to doing..

Thanks for clearing all that up ☺

PR #110, opened 2018-03-07
Fix #90: Negative scales

### LINKED ISSUE #90: Mirror not working with Transform option
Mirror feature can be applied to Transform or Geometry.

It does work properly for Geometry but not for Transform.
The mirrored mesh (with Transform option) must be ResetXForm + normal flip to make it work.

mirrors
mirrors2
Left to right: Original, Mirrored Transform, Mirrored Geometry

Max file + babylon file:
[Mirrors.zip](https://github.com/BabylonJS/Exporters/files/1695747/Mirrors.zip)

tkazik:
Fix for https://github.com/BabylonJS/Exporters/issues/90
Use babylon decomposition instead of 3ds max built-in values and adapt normals/winding order accordingly.


noalak:
Wouldn't be simplier to use the following rather than BabylonMatrix.FromValuesToRef?
```
var tm_babylon = new BabylonMatrix();
tm_babylon.m = localMatrix.ToArray();
```
Some cleanup:
- parityLocal & parityWorld are not used (Mesh line 334)
- erase old stuff (Mesh line 889)
- remove tm_babylon_test (Mesh line 929)

Can you clarify cesium treatment ? (Mesh line 587)

Can you add a link to the 3ds max doc about parity and negative determinant? (Mesh line 585)

I'm not surprised 3ds max api is bugged for decomposing matrix.

I need to double check the need of flipping normals (normal *= -1) **and** reversing winding order (0, 1, 2 => 0, 2, 1)). Seems like doing the same thing twice but at different level. Probably something i don't understand here.

I think decomposing matrix through Babylon api is required for camera and light too. Not necessary in this PR though. Just something i would like to talk about.

tkazik:
Thank you very much for your feedback!
Regarding your suggestions:
- Sure, using `ToArray()` is much nicer
- Will get rid of all unnecessary comments and unused testing variables

I also had the impression, that flipping the normals and reversing the winding order is kind of the same thing (Note: I am not an expert in 3d objects/meshes at all). But I tried lots of different options based on the parity of the 3 matrices (object, world and local) and the current implementation is the only one giving correct results. There is lots of reading material in three.js, see [one](https://github.com/mrdoob/three.js/issues/4904), [two](https://github.com/mrdoob/three.js/pull/12720) and [three](https://github.com/mrdoob/three.js/pull/12787).

Sure, [here](http://help.autodesk.com/view/3DSMAX/2018/ENU/?guid=__cpp_ref_class_matrix3_html) is the link to the documentation of the parity: Look for the last public member function. Screenshot of offline docs:
parity


If you need a testing model, this one is perfect, as it has lots of tricky parts:
[A319GEN_onlyEngine.zip](https://github.com/BabylonJS/Exporters/files/1799197/A319GEN_onlyEngine.zip)

Regarding the Cesium treatment: Currently, Cesium is not expecting a reversed winding order for negatively scaled objects (I filed a bug report [here](https://github.com/AnalyticalGraphicsInc/cesium/issues/6304)). So replacing line 591 with 588 of the file BabylonExporter.Mesh.cs is a current workaround to get the models displayed correctly in Cesium (until the issue is fixed).

Good catch with the camera and light. These will have to be updated as well. Although I think it is not really likely that these are mirrored.

One last think: I noticed that the babylon decomposition sometime returns a quaternion that have not unit length (e.g. 0.9996), which is enough for the gltf-validator to complain (`ACCESSOR_NON_UNIT`). Therefore, I now added a normalization to the quaternion after the decomposition.

noalak:
Tested and approved.

Modifications on .csproj, .sln and Designer.cs are intended? If not could you remove them? Merging afterwards.

tkazik:
Thank you very much!
Now the PR contains only the two relevant updated files....sorry for the hassle.

noalak:
Thank YOU for your work!

PR #117, opened 2018-03-19
allow overwriting existing bitmaps when saving bitmaps

Selmar:
this was the default for every other file open I could find in the project (and is quite convenient)

As with the other pull request, could not test with the latest version (I've tested with an older version and it worked fine)


noalak:
Hi Selmar,

Just to understand, the error you got when trying to overwrite a bitmap is from GDI+? 

I'm using Windows 10 and the exporter is able to overwrite existing bitmaps. Which version of Windows are you using?

Selmar:
Windows 10. There was no error at all. It simply did not replace the existing bitmap.. I have no idea why, but this fixed the problem for us.

Selmar:
Perhaps it's a 3dsmax issue? We do use different versions if I recall correctly.

noalak:
Just tested on 3ds max 2015, bitmaps are replaced. Would be nice to know why it doesn't work for you.

Anyway, PR tested and approved.

Selmar:
I really have no clue. It is a problem on all computers in the company.

We do use max 2016, not 2015. Other than that, perhaps some strange user permission settings? However, other applications don't seem to have similar problems.

PR #161, opened 2018-04-24
[Max] Animation group tool

Selmar:
From #7.

Remaining issues (most likely an incomplete list):
- Babylon exporter part not done.
- Animations will not be correct if the keyframes are linear and the start/end times are not exactly on the frames being exported. In this case, it will not interpolate (todo). For non-linear keyframes this does not happen, because the curve is transformed into a series of keyframes instead.
- Option for exporting non-animated nodes inserts an animation key, but this currently does not correctly take the pivot into account (I think this is an issue with the exporter in general, actually).

I have some time to help integration, of course, if you require any specific help don't hesitate to ask.


noalak:
Hi Selmar,

Sorry for late review!

YOU'VE DONE A VERY GOOD JOB HERE!

Tested with a simple animated box and a rigged humanoid. All good!
Find [attached](https://github.com/BabylonJS/Exporters/files/2032208/Samples.zip) the sample files (3ds Max 2017) and the glTF ouputs.

@deltakosh  Here is a screenshot of the UI:

screenshotui

### About the issues you reported:

1. Babylon not done. Np!

2. Linear interpolation not done. Understood.
@deltakosh The issue is that animations optimsation remove some keys, potentially a lot, and the desired start or end of an animation group is missing.
Do we skip that issue and raise a warning when facing such case or do we implement the linear interpolaton right now ?
If it's a warning, is it an advice "You should disable optimzation..." that let the user manage the output ?
Or is it to inform the user that optimization has been skipped whatever is the actual value, like "Animations are NOT optimized when animation groups are defined"?

3. Scale animations of bones has been fixed recently. Should fix that.

Next are my comments (AG = Animation Groups)

### About Integration:

Delete the "_removeScale" from GLTFExporter.Skin. Should be done automatically when updating PR.

### Critical issue:

Simple animations are not exported in glTF. By simple I mean no AG at all, only the standard animations defined in 3ds Max. This is a design question that needs to be adressed before merging.
@deltakosh Do we export an extra AG containing all animations as it is now ? Do we exclude nodes being part of an AG defined in the AG window ?

### Major issue:

Export process is quite long. Rigged elf takes 18s to export on my computer instead of 3s. Is it only because of the Loader.Core.RootNode.FindChildNode calls ?

### Minor issues:

When changing the scene hierarchy while the AG window is opened, the animation nodes (nodeHandles) are updated only when another setting is updated. This is the reason why you added the warning note ? Any other reason for this note ?
Can the "undefined behavior" be bypassed by closing and opening the window again ?

When creating a new scene, animation nodes are cleared but groups are not. No error on export though.

An error occurs when trying to display AG window while already visible.

### Extra features to think about:

Check that AG boundaries are not absurd (timeline.start <= start <= end <= timeline.end)

Press TAB to switch between "name", "start" and "end" input fields

Add a minimal limit on window size

### Questions/Discussion:

You serialize the data to store them between work session and scene switch, right ? If so, that's brillant!

Why is submit required ? From a user POV I think it could be annoying. What does your team say about it ?

Why would I export animations of non-animated nodes inside node tree ? Rewording, why export dummy/fake scale animations if none are present ?

And the last but not least, animation groups are clamped between Start and End frames (49 to 73 in the Jump example). The question to normalize each AG from t=0 arised lately. By that I mean that the Jump AG would start from 0 instead and wait for 49 frames before actually animating the model for 25 frames. Is it something you would desire too ?
Here are the discussions about it: [Khronos issue](https://github.com/KhronosGroup/glTF/issues/1327) and the related [Babylon issue](https://github.com/BabylonJS/Babylon.js/issues/4224).

Don't missunderstand the big blob of text, you definitely did a great job, THANK YOU !

deltakosh:
2. We should just make sure that start and end keys are exported. This is an update for the animation optimizer

If there is no AG then we should export everything like right now (One big AG)
If there is at least one AG, then we do not export the big one

Congrats @Selmar on the huge work!

Selmar:
Hey! Thanks for the comments! I'm glad you are happy with the result so far.

I will address the questions I can answer and express some concerns I have.

#### Animation optimization
A few problems were encountered:
  - We created a .gltf file that just exports a single node (without mesh), to which a camera is attached in our game. Due to the optimization, the movement became very jaggy and ugly.
  - We have large models (think 50+ meters), where some parts have relatively tiny animations (buttons that can be pressed). Animation optimizations caused the button animations to be discarded completely. I believe that, even with optimizations, at least the start- and end- keyframes should be set to their actual value and not "optimized".

I believe it should eventually at least be possible to specify optimization per animation group. Perhaps by default ON, but can be turned off? Or perhaps per node/channel? I'm not sure.

#### Exporting non-animated nodes:
The reason for this option is that we target the nodes inside the animation groups for gameplay purposes. In some cases, we need to target a (group of) node(s) that lacks any animations in the .gltf. Our main use cases for this are hiding parts of the model and highlighting (groups of) meshes. The option allows artists to not have to add dummy animations manually in this case. I realize this is rather specific, but glTF doesn't support any other way of arbitrarily grouping nodes.

The option could also be used to "force" a certain state of a node.

From our artists' point of view: if a node is selected as part of an animation group, why would it NOT be a target of the animation, regardless of whether it actually has an animation or not? If it doesn't have an animation, it can simply be deselected. The artists argue that exporting it should actually be the default.

Bugs: 
 - it does not export for leaf nodes regardless of whether this option is ticked. In fact, I think this is unrelated, I think leaf nodes without meshes are perhaps not exported at all?
 - The animation state used is taken from frame 0 I believe, but should probably be the first frame of the AG.

#### Long export process
I haven't looked into this at all. I am aware the export process is not very fast, but we have some very large models, so that was quite normal for us to begin with.

#### Editing scene hierarchy & AG window
The "undefined behaviour" warning is indeed because the node view is not updated dynamically. Deleted nodes are removed on refresh only. This also triggers a re-serialization, I believe. It can indeed by bypassed by closing and re-opening the window after editing the scene view.

I'm not sure if the serialization is entirely correct, judging from new scene creation. Or perhaps the handle to the root node is simply a handle to the "old" scene? I don't know. I believe I copied the way that works from the scene-tied serialization that you already had, so any bugs related to that might also exist with the existing options you have per-scene.

There is one issue that one of the artists reported that worries me; there was one instance where he lost his work. I know he had several 3ds max instances open, but I'm not sure if that was the cause. I have not been able to reproduce any loss of work and thus I have not found the root cause of the problem.

#### Confirm on AG changes
This option is there mostly because I don't want to serialize every time the user changes anything. I felt like it could become slow for large models, as it is mostly string-based. There is an added benefit: it is not so easy to accidentally change something and not know what it was. This adds a layer of security for the artists' workflow. So I implemented it like this, per AG, with colors that show what changed so that it is clear something has not been saved.

The artists here are very positive about the tool and don't have any problems with the way it works. I have the impression they like the explicitness of the confirm button.

#### Normalizing animation groups
As I started replying to this, I realized that you probably meant the way the viewer treats this data and not actually change the data? For the viewer, we would prefer non-normalized animations (e.g. just start where the animation starts). We have a large timeline with a lot of small animations all over the place. However, we don't use it much beyond the initial modeling stage, because we have a few extensions (e.g. custom material) and a different way of organizing textures.

From a .gltf point of view it could make sense to want an animation to start at t=0, but in my view that means it should then end at t=24. However, in practice, it is not so hard to ignore the start time and only consider the duration when playing an animation. If you want an animation to have a 49-frame delay, I think it should just have 49 extra frames representing this delay.

Explicitly inserting a pause/delay seems out of scope of what the glTF animations represent, and is something that I think should be part of a client or animation sequencer extension/feature.

deltakosh:
> the start- and end- keyframes should be set to their actual value and not "optimized".
Yes :)

> Normalizing animation groups
If animation is expected to start immediately then we just need to offset it to 0. Meaning that an AG specified from 20 to 40 will be stored in glTF from frame 0 to frame 20 (using the right animation data of course). This is what the spec dictates

Selmar:
Normalizing that way is fine for us and should just work. :)

De : David Catuhe <notifications@github.com>
Envoyé : Thursday, May 24, 2018 16:11
À : BabylonJS/Exporters <Exporters@noreply.github.com>
Cc : Selmar <selmar21@hotmail.com>; Mention <mention@noreply.github.com>
Objet : Re: [BabylonJS/Exporters] [Max] Animation group tool (#161)


the start- and end- keyframes should be set to their actual value and not "optimized".
Yes :)

Normalizing animation groups
If animation is expected to start immediately then we just need to offset it to 0. Meaning that an AG specified from 20 to 40 will be stored in glTF from frame 0 to frame 20 (using the right animation data of course). This is what the spec dictates

deltakosh:
Sounds good!! 
@noalak let's do a merge as soon as the normalization is done
@Selmar and @noalak do you guys can start a new issue to list remaining tasks / bugs?

noalak:
Ok I will create a new issue and Selmar will comment on it afterwards.
I think we should handle non-AG animations too before merging.

deltakosh:
sounds good to me

noalak:
### Animation optimization

> We created a .gltf file that just exports a single node (without mesh), to which a camera is attached in our game. Due to the optimization, the movement became very jaggy and ugly.

Could you open a new issue with a repro scene ?

> Animation optimizations caused the button animations to be discarded completely.

Animation optimizations can currently be turned off at global level (scene properties), default is on. Did you know that ?

>I believe it should eventually at least be possible to specify optimization per animation group. Perhaps by default ON, but can be turned off? Or perhaps per node/channel? I'm not sure.

Still think an option per AG should be available ?


### Exporting non-animated nodes

>The reason for this option is that we target the nodes inside the animation groups for gameplay purposes. 

Understood and indeed seems pretty useful! Kind of out of the scope of glTF exporter but I see no problem keeping it as an option, false by default. Will be documented.
To go even further, we could imagine the AG system being used even without any animations, to act like layers in 3ds max.

>it does not export for leaf nodes regardless of whether this option is ticked. In fact, I think this is unrelated, I think leaf nodes without meshes are perhaps not exported at all?

Indeed, only mesh, light and camera types are considered relevant to export. Other types are exported as empty nodes only when necessary from a hierarchy POV (parent of an exported mesh/light/camera). It is possible to include more type as relevant, like helpers, but that would be an option. Skinned mesh may use helpers (dummies) as skeleton instead of real bones and would create a mess in the output scene.

>The animation state used is taken from frame 0 I believe, but should probably be the first frame of the AG.

Since the node is not animated all local TRS values remain unchanged, so any frame could do the trick even one outside timeline boundaries.


### Editing scene hierarchy & AG window

>I'm not sure if the serialization is entirely correct, judging from new scene creation.

We will fix issues as they come once merged using bugged scene to repro.


### Confirm on AG changes

Make sense!

noalak:
Things to do before merging:

- [ ] Update Babylon animations optimization to keep all AG start and end frames
- [ ] Export a single AG targeting all animated nodes when no AG is defnied in the AG window
- [ ] Offset AG to start at 0

@Selmar Do you have some free time for this ? I can help you out if needed!

Selmar:
I think I may be able to find some time next week or the week after, but I don’t have a lot of time unfortunately.

I skipped the animation optimization part because it seemed part of the Babylon parts (which I didn’t touch), so perhaps you could tackle that one at least.

De : Tambrun Fabrice <notifications@github.com>
Envoyé : Friday, May 25, 2018 15:22
À : BabylonJS/Exporters <Exporters@noreply.github.com>
Cc : Selmar <selmar21@hotmail.com>; Mention <mention@noreply.github.com>
Objet : Re: [BabylonJS/Exporters] [Max] Animation group tool (#161)


Things to do before merging:

  *   [ ] Update Babylon animations optimization to keep all AG start and end frames
  *   [ ] Export a single AG targeting all animated nodes when no AG is defnied in the AG window
  *   [ ] Offset AG to start at 0

@Selmar<https://github.com/Selmar> Do you have some free time for this ? I can help you out if needed!

Selmar:
This is a bit of a sidetrack, but what about nodes that are normally ignored? If we target a node with an animation group, but that node is a leaf node without mesh/light/camera, we might want some feedback for the user so that he knows why a node might not be exported. In fact, we often want to export nodes for gameplay purposes (attaching things), which also happens outside of animation groups.

The workflow for us is usually that helpers and such go in a separate layer which is hidden on export, after which we export non-hidden objects from the scene. In that case, it makes a lot of sense to always export everything that is visible. I think, at least for us, an option for that would be very useful. Artists are regularly confused why certain things are not in the .gltf file when they were selected for exportation (like with the animation group nodes).

Thoughts?

noalak:
Nice you have some time for this!

About animation range check, I think we should raise warnings when AG range values are corrected. At the very least when AG range is wholly out of timeline range, resulting in a single value range ([X, X]).

Selmar:
Done. Also fixed a bug with the range check, because I forgot that AnimRange is in ticks and not frames. It's not changing the animation group values themselves, I feel like that is the responsibility of the artist himself.

noalak:
Pending PR to optimize animations.
Checking last features before merging.

noalak:
All good for me.

Ok for you @Selmar?

Selmar:
Yeah, looks good.

I'll be testing it internally after the other issue has been fixed.

PR #179, opened 2018-05-08
Update mesh.py

babbleon:
Adds basic support for mesh constraint saved under each mesh as metadata.


Palmer-JC:
Will add a separate commit 2 update the version #, & create the zip distrib file in a day or so

Issue #160, opened 2018-04-24
[Max][glTF] Exporting a skin can lead to exporting an incomplete hierarchy.


Selmar:
Exporting a skin with bones that form an incomplete hierarchy result in an incomplete hierarchy in the .gltf file.

You can see below that the node with index 4 is `Dummy002`, which should be parented to `Dummy001` and not to the scene root.

![image](https://user-images.githubusercontent.com/3010534/39199706-6d63766a-47ea-11e8-94c2-5a79aa6bf1f4.png)

![image](https://user-images.githubusercontent.com/3010534/39199758-86cf377e-47ea-11e8-8801-25ba52c12ec5.png)

Selmar:
Note that removing data duplication (#159) would also fix this issue!

Issue #159, opened 2018-04-24
[Max][glTF] Skin exports lead to data duplication


Selmar:
I believe this originates from the fact that there is a separation between nodes and bones in the babylon format, which does not exist in glTF.

For most (all?) skins, a separate hierarchy is exported. It would be nice to avoid this, as the number of nodes and with it amount of animation data grows quickly with larger models.

deltakosh:
You mean:adding a checkbox to not export bones node?

Selmar:
From my perspective it's simply reducing redundancy in the exported data. I don't think this should require a checkbox. It should simply not export a duplicate hierarchy (it doesn't exist in 3dsmax either). 

There's also #160, the hierarchy may not even be correct right now, partially because of the bone/node duplication.

If skeleton duplication is needed I feel like that should be handled client-side, not in the exporter.

Selmar:
The issue with partial hierarchies is fixed and the amount of redundancy has been reduced, however, every skin still exports a separate hierarchy.

The code adds some checks to detect "identical" skeletons, which reduces some duplication. However, glTF does not have a strict representation of a skeleton (because it does not have bone modifiers, unlike 3dsmax), it only has a skin that references existing nodes (which is how 3dsmax skins work, too). So, from my point of view, if a node is referenced by a skin, its representation in the hierarchy should be referenced by that skin instead of duplicated.

I have attached an example. `Dummy001`, `Dummy002` and `Dummy003` are duplicated in the exported hierarchy.

[testAnim.zip](https://github.com/BabylonJS/Exporters/files/2045546/testAnim.zip)

Selmar:
I would like to re-open this isuse, as there is still (or again?) a lot of duplication in the skinning data. Observe the below output from one of our bigger meshes.

Many joint hierarchies share the same `skeleton` root (which is obsolete, but it gives us some information now). They often share a large part of the hierarchy. In some cases, the hierarchy is exactly the same (`#18`, `#18_1` for example, and `#19`, `#19_1` and `#21`).

This seems redundant?

```json
    "skins": [
        {
            "inverseBindMatrices": 2026,
            "joints": [
                63,
                71,
                72,
                73,
                74
            ],
            "name": "skeleton #0",
            "skeleton": 63
        },
        {
            "inverseBindMatrices": 2027,
            "joints": [
                96,
                105,
                106
            ],
            "name": "skeleton #1",
            "skeleton": 96
        },
        {
[...]

PR #193, opened 2018-05-24
[Exp - Maya] Skin

rlouvat:
Maya skeleton:
- Optimization
- Fix duplicated nodes in glTF
- Fix duplicated skeleton in Babylon export
- Add progress bar update for the skin export
- Add comments in the code

3DS Max:
I found a way to fix the duplicated skeleton bones issue #159 .
But I need a full hierarchy... So it will work after the fix of #160  (I'm working on it). 

@deltakosh : is it ok for you ?


deltakosh:
No problem!

Issue #198, opened 2018-05-25
Blender: No export options


tristan-shelton:
In both Blender 2.79 and 2.8 preview I'm noticing that the Blender exporter doesn't show any export options at all. It seems to be a bug or omission in the official exporter which while great to have seems pretty limited.

I patched the exporter to serve my own purposes for the moment, but I'm curious if there's any interest in pull requests for the exporter? I'm no Python expert but I think I could improve what's currently available, at least a little bit.

deltakosh:
Pinging @palmer_jc

deltakosh:
But anyway all PRs are welcome obviously

tristan-shelton:
Great! I'll patch up what I changed and put in a small initial PR. :smile:

Vinc3r:
JCPalmer commit new version with this [commit](https://github.com/BabylonJS/Exporters/commit/4da6f638e4a02d423910ab964a4cde3609983e8e), I think this issue can be closed isn'it?

PR #199, opened 2018-05-25
Update blender exporter settings

puck7744:
- Added seemingly missing mixin class for settings panel
- Removed the export active layers option
- Added an export scope option
  - "All" option to export everything in the Blender scene, similar to previous functionality
  - "Selected" option to export only selected objects
  - "Visible" option which replaces the active layer option and also works with Blender's hide/show feature

### DIFF
--- a/Blender/src/babylon-js/__init__.py
+++ b/Blender/src/babylon-js/__init__.py
@@ -45,6 +45,7 @@ else:
 
 import bpy
 from bpy_extras.io_utils import ExportHelper, ImportHelper
+from .exporter_settings_panel import ExporterSettingsPanel
 #===============================================================================
 def register():
     bpy.utils.register_module(__name__)
@@ -64,7 +65,7 @@ if __name__ == '__main__':
     unregister()
     register()
 #===============================================================================
-class JsonMain(bpy.types.Operator, ExportHelper):
+class JsonMain(bpy.types.Operator, ExportHelper, ExporterSettingsPanel):
     bl_idname = 'bjs.main'
     bl_label = 'Export Babylon.js scene' # used on the label of the actual 'save' button
     bl_options = {'REGISTER', 'UNDO'}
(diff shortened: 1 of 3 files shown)


Palmer-JC:
Per the forum thread, the default place with labels pointing to location preserves savability.
` See Scene Properties`
`For Savable Exporter`
`Properities.`

The expansion of selection is an improvement, but think the active layers still needs to be added the list.  Reason is that hide/show is used to export things that import which are also not initially displayed.

puck7744:
Ah, that makes sense! I'll see if I can update the request. :)

Vinc3r:
Thanks for this PR, it's a nice improvement :)

deltakosh:
@Palmer-JC  ok to merge?

puck7744:
I updated the request to fix some issues with the way I had changed the exporter in line with what @Palmer-JC said.

I also added a simple message to the export settings panel so that it's clear that the lack of ssettings when exporting is not a bug.

Finally, just for my own sanity I changed the `bl_idname` variable to fit Blender's format and thus allow a shortcut to be set for the exporter.

puck7744:
If you do decide to merge, feel free to close #198 as I think it's quite well taken care of 😃

Palmer-JC:
Done for the day, will look at this tomorrow

Palmer-JC:
Will generate a zip, after upping version number, & nuking the unused isInSelectedLayer function.  Probably today.  Thanks!

tristan-shelton:
@Palmer-JC I actually did re-use isInSelectedLayer here: https://github.com/BabylonJS/Exporters/pull/199/files#diff-dc6dd5305e467a57d4953ed20abe6f7dR85

Looking at that section again I also see that I made a mistake in that it should be operating on `scene.objects` not visible objects. You'll probably want to fix that up before making a release. Sorry about that!

To be clear, line 86 should be `scene.objects` not `bpy.context.visible_objects`

Issue #197, opened 2018-05-25
Animation groups enhancement


noalak:
Listing enhancements to [animation groups](https://github.com/BabylonJS/Exporters/issues/7) following [Selmar work](https://github.com/BabylonJS/Exporters/pull/161/).

### Major

- [x] Documentation
- [x] Export animation groups to Babylon too and convert them to glTF afterwards
- [ ] Optimize export duration (see Loader.Core.RootNode.FindChildNode)
- [ ] Update the node tree view dynamically when scene is changed while AG window is opened

### Minor

- [x] Check that AG boundaries are not absurd (timeline.start <= start <= end <= timeline.end)
- [x] Do not try open AG window when already visible
- [ ] Clear AG when creating a new scene
- [x] Press TAB to switch between "name", "start" and "end" input fields
- [x] Add a minimal limit on window size

Selmar:
I can't open the RiggedElf example from #161, so I've profiled on a different model. This leads me to believe that the `Loader.Core.RootNode.FindChildNode(nodeHandle);` is definitely not the bottleneck. The export was approximately 30 seconds in total.

The model has around 400k triangles, is mostly editable polys without skin. For the export, I had created 10 animation groups and added all 58 nodes to every animation, with the "export non-animated nodes" option selected.

Profiling was done in Release (I believe you might be using debug builds, though?)

![image](https://user-images.githubusercontent.com/3010534/40613558-41d2d22a-627f-11e8-97ed-60288ebab5a5.png)

Selmar:
Just verifying, isn't the interpolation fix missing? I.e. exporting the correct data for the start- and end-frames of an animation? (because right now it might be incorrect if there is no keyframe on it).

Selmar:
About tabs: I can't get it to work. I had already set some tab indices on the controls, but it doesn't seem to do anything. It doesn't work on the other babylon forms either.

Selmar:
Done in #161: 
- a bug where animation data was referenced instead of copied before doing the axis conversion, leading to a double conversion in some cases
- export all animations in one animation group if there are no animations
- clamp the range of the animation group frames before using the values

Selmar:
The exception on opening the window twice was due to it being a modal window, in which case calling Show() causes an exception. Fixed. 

This seems like something convenient to have for all windows, though, as it makes the form stay in front of the main window? Is that wanted or not? Currently it's inconsistent with the exporter window.

Selmar:
[...]

PR #201, opened 2018-05-28
3DS Max: Update => from now on Helpers (Dummy) are always exported

rlouvat:
As requested by @Selmar in #161 

@deltakosh is it ok for you ?


Selmar:
I have a question: the `alreadyExportedNodes` array uses the `name` property, but the `name` is not guaranteed to be unique for a single node, is it? Shouldn't it use the node.id or something like that? @noalak  ?

Selmar:
I also think this does not work with the AnimationGroup tool and has to be revised before we merge that. The animation groups export node/bone animations in a different place. @noalak #197

rlouvat:
Thank you Selmar for your pertinent observation. Instead of the name (I though they were unique), I will use the id.

Selmar:
Another issue: it is exporting helpers that aren't selected if "export only selected" is ticked.

rlouvat:
In this pull request #204 , as suggested the `alreadyExportedNodes` uses the `id` property and the issue is fixed.

PR #218, opened 2018-06-06
[Exp - Max - glTF] Fix the node export when a mesh is a bone child

rlouvat:
It fixes a side effect of #159.
When a mesh is a child of dummies that are used as bones, this mesh was exported as a simple node (its mesh properties weren't exported).

---

Unfortunatly, there is one specific case where I can't find a solution: when the mesh uses its own parents (dummies) as bones.
The example files [testAnim.zip](https://github.com/BabylonJS/Exporters/files/2076360/testAnim.zip)

In this configuration with the export to babylonjs:
- The mesh has twice the normal animation: one due to its parent (the bone) and one due to its skin (the same bone again).
(In the example files, the blue cube rotation on the vertical axis is (-90° * 2 =) -180°)

In this configuration with the export to glTF:
- The .gltf file contains a correct hierarchy (the mesh is the child of its own bone).
After loading it in the BabylonJS Sandbox the mesh has no parent but the animation is correct.
(Using the example files, the blue cube has `__root__` as parent and its rotation on the vertical axis is -90°)
@deltakosh  Is this hierarchy modification the expected behavior?

You can notice that in the ThreeJS viewer the animation is the same as the one in the BabylonJS Sandbox. So it should modify the hierarchy the same way.


deltakosh:
This is ok so far.let's keep it like that for now

PR #221, opened 2018-06-07
[ Exp Maya glTF ] Fix the node export

rlouvat:
Pushing the modifications done on 3DS Max to Maya (see #218 #159).

---

Unfortunately, the specific case where the mesh uses its own parents as bones is not working...
But unlike 3DS Max, in this configuration the mesh has a right animation in babylonjs and a wrong one in glTF.


deltakosh:
Not a big deal

PR #215, opened 2018-06-05
[Exp - Max] Export tangents

rlouvat:
3DS Max:
- Add tangent export for babylonjs and glTF (there is a checkbox to export or not the tangents) #60 .

PR #222, opened 2018-06-08
[3DS Max] Tangents

rlouvat:
3DS Max:
- add tangent export for the morph target
- the tangent w is computed instead of a fix value

In addition of #215.
@deltakosh My bad I wasn't aware of the morph's target and I have discovered that flipping the normal change the value of the tangent w.


deltakosh:
no problem!

Issue #194, opened 2018-05-24
[max][gltf] non-uniform scales result in distorted normals


tkazik:
How to reproduce:
Create an object, scale it with different scales along the three axis (node scales). The exported gltf seems to have distorted normals compared to the same object that has uniform scales (with resetX), see:

![appearance_babylon](https://user-images.githubusercontent.com/15689124/40514279-d9d20368-5fa8-11e8-9d10-a73b77b87336.png)

I suspect that we are missing something during the export...some sort of scaling of the normals is missing. Here you see the normals in max:

![non-uniform_scales](https://user-images.githubusercontent.com/15689124/40514271-d730d4ae-5fa8-11e8-8460-634ace62b0d3.png)

Here is the sample model including the glb:
[nonUniformScales.zip](https://github.com/BabylonJS/Exporters/files/2037143/nonUniformScales.zip)

Any ideas? Thx!

deltakosh:
Can you try to bake transforms into vertices before exporting?

deltakosh:
OK makes sense
I'll have a look asap

deltakosh:
Ok it will be fixed by next commit
Now you can do this:
```
mesh.ignoreNonUniformScaling = true;
```

@noalak: Do you mind exporting all 3dsmax meshes (in Babylon.js) with this option set to on?

tkazik:
Puh, you were quick! 
Just to double-check: You fixed it in the framework, right?
My gut feeling tells me, that there might be something fishy with the exporter, as also three.js displays the model incorrectly. So either, the normals are messed up during export or three.js needs the same fix...what do you think?

deltakosh:
It is a framework fix and yeah three will probably need some way to disable non uniform scaling process for normals

tkazik:
@noalak: Do you see an option to adapt the normals in the exporter? This would be an alternative...instead of adapting all loaders. Or what do you think?

Selmar:
The normals are currently retrieved in world space. I remember there was a problem retrieving this information in local space, I do not remember why...
`Normal = mesh.GetNormal((int)face.Norm[facePart], false)`

After that, they are transformed into local space and normalized:
`vertex.Normal = invertedWorldMatrix.VectorTransform(vertex.Normal).Normalize;`

Checking the output, everything seems to be okay actually. Here's a snippet from the raw data:

```
one mesh:
   -0.01375     0.99953    -0.02750 
   -0.01009     0.98884    -0.14863 
[...]

Issue #214, opened 2018-06-04
[max][gltf] non-uniform objectoffsetscales result in distorted normals


tkazik:
Very similar to https://github.com/BabylonJS/Exporters/issues/194 .

The above (solved) was for the node scales, this issue is for the object-offset-scales. For background info see:

http://help.autodesk.com/view/3DSMAX/2018/ENU/?guid=__files_GUID_3B001F21_8FE9_4663_A972_E648682A0ACD_htm
http://help.autodesk.com/view/3DSMAX/2018/ENU/?guid=__files_GUID_C17D0E4B_C408_426B_B300_7E23A0D98515_htm

Or in maxscript, type:

`$.scale` vs. `$.objectoffsetscale`

Sample screenshot of the glb-model:
![fixobjoffsetscales](https://user-images.githubusercontent.com/15689124/40943686-5ba80a52-6852-11e8-960f-820af6d3921b.png)

The model: [nonUniformObjectoffsetscales.zip](https://github.com/BabylonJS/Exporters/files/2070098/nonUniformObjectoffsetscales.zip)

Thx a lot!

noalak:
Hi tkazik,

Thanks for reporting another issue. Looks like you are already well informed!
You've contribute to the project before, maybe you can keep up the good work now :) ?

Vertex normals are retreived from 3ds max in the 'CreateGlobalVertex' function of the BabylonExporter.Mesh.cs file.

tkazik:
Sure, I can have a look...just thought you might be way quicker, as you probably have a much better understanding of both 3ds max and the c#-code.
Will keep you updated.

deltakosh:
Any news?

tkazik:
Still working/digging on it...time is a bit sparse at the moment :)
The interesting scales of the offset transform are hidden under the following:

`meshNode.MaxNode.ObjOffsetScale.S`

So I just need to figure out, how to apply them properly. Hope to be able to create a PR soon.

PR #229, opened 2018-06-18
Non uniform object offset scales

tkazik:
suggested fix for https://github.com/BabylonJS/Exporters/issues/214
(also fixes https://github.com/BabylonJS/Exporters/issues/194 for negative scales)

old vs new:
old_vs_new

Again, few hours of debugging needed to sort of understand what 3ds max does under the hood (unfortunately, their sdk documentation is not really complete and it could be much more comprehensive). The code should be quite self-explanatory, but here is a short wrap-up:

1. The interesting variables (`ObjOffsetScale` and `ObjOffsetRot`) are properties of `meshNode`, which is why `meshNode` is propagated all the way to the function `CreateGlobalVertex`.

2. The objectOffset-scale and -rotation are in a LH coordinate system.

3. Actual computation of normal: Undo offset rotation, apply offset scale, re-rotate normal

Note: To deal with negative node and offset scales, I had to take the absolute value during the normal computation. Probably, there is a cleverer way to do that (connected with the parity check), but I wasn't able to figure that out...and the current version runs fine.

To do / future work / open points:
- Needed: Move tangent computation after normal computation, as the tangents depend on the normals
- Nice to have: Think about propagation of `meshnode `and `babylonAbstractMesh`...maybe there is a nicer way.

- If plenty of time available: There should be a way to get rid of the absolute value function during normal computation...might be related to the parity checking and normal flipping.

Looking forward to your feedback

PS: Some testing material: [testModels.zip](https://github.com/BabylonJS/Exporters/files/2113157/testModels.zip)


noalak:
Good job!

noalak:
Doing a PR to update dlls.

PR #236, opened 2018-06-25
[Maya - glTF] KHR texture transform

rlouvat:
Maya - glTF:
Add KHR_texture_transform extension.
Regarding the texture offset, rotation and scale, the .glTF export have the same result in the sandbox as the .babylon export.


deltakosh:
Good job!

PR #250, opened 2018-07-03
Use correct parent when exporting meshes. Fixes #247

### LINKED ISSUE #247: [Maya] Instances of groups are exported to incorrect positions
When creating an instance copy of a group of objects the instanced objects are exported to the same position as the original object even after moving the group to a different location.

**Steps to reproduce:**
1. Create some object (for example a cube)
2. Group the object (Edit -> Group; Ctrl+G)
3. Create an instance of that group (Edit -> Duplicate Special with geometry type set to "Instance")
4. Move the new group
5. Export the scene

**Result:**
In Maya the two cubes are at different positions. In the exported Babylon scene both cubes are at the same position (therefore it looks like only the original object has been exported).

The following zip contains an example scene and the exported Babylon scene: [instancegroupbug.zip](https://github.com/BabylonJS/Exporters/files/2150082/instancegroupbug.zip)

FrozenBrain:
Currently the exporter uses the id of the first parent of a mesh as its parentId when exporting it. This results in meshes having an incorrect parent and therefore incorrect positions and rotations in the scene.

In the example scene I attached to issue #247 the Outliner shows the correct object hierarchy:
image
The exporter however exports both cubes as children of group1:
image

After the changes in this PR both cubes are exported and grouped correctly:
image

### DIFF
--- a/Maya/Exporter/BabylonExporter.Node.cs
+++ b/Maya/Exporter/BabylonExporter.Node.cs
@@ -45,7 +45,10 @@ namespace Maya2Babylon
             {
                 RaiseVerbose("BabylonExporter.Node | Hierarchy", 2);
 
-                MObject parentMObject = mFnTransform.parent(0);
+                var mDagPath = new MDagPath(mFnTransform.dagPath);
+                mDagPath.pop();
+
+                MObject parentMObject = mDagPath.node;
                 // Children of World node don't have parent in Babylon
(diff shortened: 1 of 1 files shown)


rlouvat:
Hi,
I looked at your commit and it is all right for me.

noalak:
Updating dlls right now

noalak:
@FrozenBrain Thanks for your work!

PR #253, opened 2018-07-04
[Exp Maya] Morph target

rlouvat:
Add morph target export for maya:
- export to babylon and glTF
- influence animation
- add sample like the one for 3DS Max

Known issues:
- Mesh with both morph target and skin are not exported correctly.


deltakosh:
There are conflicts

PR #258, opened 2018-07-10
[Skin] vertex bone checks

Selmar:
Changes:
- ignore bones with weight <= 0
- output vertex indices for vertices with >4 bones (gltf) or >8 bones (babylon)
- add check/warning for skins without bones (previously exception)

Possible issues:
- Many duplicate messages (see below image). This was already the case for babylon, before this pull request, the difference being I added this message for glTF and added the vertex index to the ouput. This begs the question: is there a reason for processing the same vertex index multiple times?

Already existing issues, even before this pull request:
- I can't get the skinned animation to play in babylon.js (the one that comes with "export & run), but it works correctly in three.js and the [online babylon viewer](https://sandbox.babylonjs.com/).
- The cone is invisible in three.js, but not in babylon!

image

max file (max2016) & gltf:
[skin-stuff.zip](https://github.com/BabylonJS/Exporters/files/2180031/skin-stuff.zip)

### DIFF
--- a/3ds Max/Max2Babylon/Exporter/BabylonExporter.GLTFExporter.Mesh.cs
+++ b/3ds Max/Max2Babylon/Exporter/BabylonExporter.GLTFExporter.Mesh.cs
@@ -374,11 +374,6 @@ namespace Max2Babylon
                     accessorWeights.count = globalVerticesSubMesh.Count;
                 }
 
-                if (hasBonesExtra)
-                {
-                    RaiseWarning("Too many bones influences per vertex. glTF only support up to 4 bones influences per vertex. The result may not be as expected.", 3);
-                }
-
                 // Morph targets positions and normals
                 if (babylonMorphTargetManager != null)
                 {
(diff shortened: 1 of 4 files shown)


Selmar:
Oh, and I see that I have created a problem. The cone is invisible. On it.

deltakosh:
We can't merge something that does not work in Babylon.js :)

Selmar:
You misread, this issue already exists right now. I just happened to notice it when I verified this pull request.

deltakosh:
Oh ok:) gotcha!
I let @noalak comment then:)

rlouvat:
Hi all,

Thank you @Selmar for the good work.

For your observation:
> I can't get the skinned animation to play in babylon.js (the one that comes with "export & run)

It's because the local server page opened by the "Export & Run" button doesn't contain the animation group footer. I will update this page by copying the one of the sandbox.

noalak:
Updating dlls

Issue #272, opened 2018-07-23
[Max] Enable/disable the default hemispheric light ?


Remi-Tribia:
Hi,
First, thanks for your incredible reactivity on the "extras" node tags!
Would it be possible to add a checkbox to enable/disable adding a default hemispheric light when exporting a scene with no light?
Cheers!

deltakosh:
So your goal will be to turn the default light off?

Remi-Tribia:
Yes.
I merge multiple gltf files on a main scene, some of them doesn't have any light. I can filter the results to get rid of it, but having the exporter adding an unwanted default light seems weird.

PR #274, opened 2018-07-24
[3DS Max] Enable/disable the default hemispheric light

rlouvat:
Following the request #272

A checkbox to enable/disable the default hemispheric light was added in the scene form.
By default the checkbox is checked (so the default light is added, if there is no light exported).


deltakosh:
There are conflicts due to the other pr I merge:)

Issue #294, opened 2018-08-14
Animation Stutter


gabrieljbaker:
When exporting a model with animations from 3DS Max, we're noticing a stutter effect on the animation when we view the exported model. See below for how this particular animation looks in Max vs. how it looks once exported with the Babylon exporter. You'll notice a subtle stutter happening. 

![ghost_loop_max2018](https://user-images.githubusercontent.com/5530378/44113823-3597ab26-9fbe-11e8-83f0-665fe2415a73.gif)
![ghost_looped_babylongltf](https://user-images.githubusercontent.com/5530378/44113829-39fc0194-9fbe-11e8-84da-59097ed21ff4.gif)

Here also is a link to the source model and exported model if that would be helpful: http://tbf.me/a/CNn2Sz

WVNicolayeff:
Artist on the project here. That top GIF from Max had a few extra frames at the end - in Max it is smooth loop.

sebavan:
Adding @noalak and @Jaskar for deeper analysis.

sebavan:
Closed by the previous PR. Big Thank to @noalak

PR #295, opened 2018-08-16
Fix issue #294

noalak:
Fix issue #294

Explanations:
Mathematically, LocalMatrix = WorldMatrix * Inverse(WorldMatrixParent)
For an unknown reason, this is not true using 3ds max API. Difference is around 1%, not huge, but definitively not a rounding error (something under 0.01% would be ok).
Using the local matrix directly seems the way to go.


tkazik:
That is weird...somehow, these transform matrices seem to be buggy (remeber the issue with negative scales etc). Would be interesting to see if the c++ sdk has the same behaviour.

PR #308, opened 2018-08-29
Minor fixes

noalak:
Export either rotation or rotationQuaternion
Store node uid inside file as before and check for undesired duplication at export


noalak:
v1.2.31 is bugged, see #309

PR #309, opened 2018-08-31
Fix critical node id issue

noalak:


noalak:
Critical bug was introduced in version 1.2.31 regarding material and instance ids

PR #312, opened 2018-09-04
[3DS Max] exporter for 3ds max 2019

innoreschke:
I added (copy and adjust from 2018) a project for exporting from 3ds max 2019.

since there were no API changes affecting the exporter only a new compile symbol had to be added and source files modified accordingly to use the correct API versions similar to 2018.


deltakosh:
LGTM
Can you check @noalak?

deltakosh:
Thanks a lot @innoreschke

Issue #296, opened 2018-08-16
Max2Babylon viewport shader?


echadwick-wayfair:
Has anyone looked into creating a viewport FX shader for 3ds Max?

It would be much faster to iterate on materials & geometry if we had a viewport shader in 3ds Max that mimics the BabylonJS default shader. If we could give this to our content creators, they would be able to iterate much more quickly than the "Export & Run" feature in the current Max2Babylon exporter. A viewport shader is a great live-editing tool.

ShaderFX is a built-in FX shader editor for Max. I've used it to create analogous Unity-style shaders for artists to preview their work directly inside Max, without having to wait for export/update in Unity.
http://ericchadwick.com/img/robot_rising.html#utility_shader

![image](https://user-images.githubusercontent.com/40177036/44220683-524afd00-a14d-11e8-9066-7940e290e831.png)

I haven't created a metalness PBR shader though, a bit beyond my skill level. 

Any interest in this?

sebavan:
Would be nice. 

The thing is with PBR at least it should be pretty close (not enough I understand :-) ) as the material definition is "pretty-ish" "standar-ish".

Would you be able to create and make a PR for it ? I can definitely help with the shader we are using.

sebavan:
I will check with our in house artists if this would be possible cause I really like the idea :-)

echadwick-wayfair:
Some implementation problems to factor in, see these 3ds Max Idea Board topics:

https://forums.autodesk.com/t5/3ds-max-ideas/pbr-viewport-shader-environment/idi-p/6806116

https://forums.autodesk.com/t5/3ds-max-ideas/native-pbr-viewport/idi-p/8071453

echadwick-wayfair:
Hey, got that to work, mostly! 

100% roughness is causing a different lighting effect on the color chart than in BabylonJS. But it's pretty darn close.

![image](https://user-images.githubusercontent.com/40177036/44233934-3ce8c980-a173-11e8-8bc9-be0b0256d9d2.png)

![image](https://user-images.githubusercontent.com/40177036/44233607-55a4af80-a172-11e8-9ae3-372ef51e782c.png)

echadwick-wayfair:
Better with a direct light.

![image](https://user-images.githubusercontent.com/40177036/44234513-c1881780-a174-11e8-8174-00ac23a4b533.png)

echadwick-wayfair:
[...]

PR #316, opened 2018-09-10
[3DS Max] Shell material

rlouvat:
The Shell material is a container for other materials and it also controls which material is used in which rendering.

The Shell material contains two materials: "Original Material" and "Baked Material". Following the discussion in #296, only the "Baked Material" is needed for the export.

---

Add the Shell material support for Babylon and glTF export:
- the Shell material itself is not exported
- its first material "Original Material" is not exported
- its second second "Baked Material" is exported
- its parameters "Viewport" and "Render" are ignored


echadwick-wayfair:
Max2Babylon-1.3.4
3ds Max 2018.4

Hmm, has this been implemented? I cannot get this material to export properly. All meshes using this material setup are exported as Dummy nodes. Here's an example to test with.

image

[box_shell-mtl_max2018.zip](https://github.com/BabylonJS/Exporters/files/2519707/box_shell-mtl_max2018.zip)

Thanks for your help with this!

echadwick-wayfair:
In the screenshot above, if I assign either the Shell Material or the DirectX Shader material to the box, the exporter treats the mesh as a dummy. But if I assign the Multi/Sub-Object material (marked with the green checkmark), then it exports properly as a mesh with vertices.

deltakosh:
Pinging @kcoley

echadwick-wayfair:
Also affects support for DirectX Shader material 
https://github.com/BabylonJS/Exporters/pull/318

kcoley:
@echadwick-wayfair I’ll take a look

PR #371, opened 2018-11-12
New toggle buttons: Add/remove default light & Cesium fix for negative scales

tkazik:
Two new options:
- the gltf-validator complains about an empty node, when the default light is added. With the checkbox, we now have the option to export models without any lights.
- Cesium currently still expects the wrong winding orders for negative scales (https://github.com/AnalyticalGraphicsInc/cesium/issues/6304). With the checkbox, we can export the model in the winding order that Cesium expects. This checkbox can be removed, once Cesium is fixed.

defaultlight_and_cesiumbug

Your thoughts? 

PS: Sorry for the commit-mess, only the last one contains the relevant changes.


kcoley:
Hi @tkazik.  The toggle for the default light looks great, though I don't think we should merge the Cesium workaround here.

Issue #368, opened 2018-11-08
[Max][glTF] Exporting .max into several glTF files.


Selmar:
Hi!

We would like to export separate scenes into one .gltf file.

I imagine it would work by somehow specifying a node as a scene root, which can optionally transform all its children to the scene root node's space (i.e. make it the origin of all its children).

This would simplify exporting many objects that are defined in the same 3dsmax file.

Remi-Tribia:
What do you call separate scenes? Separate .max files?

Cause what you are explaining is just like adding a dummy and parent all other objects to it, then export everything with the exporter.

I mainly work with Three.js (so BabylonJS might import gltf differently) but when I do that (parent dummy or grouped objects) it creates an empty Group object in which every children is parented.

Selmar:
No, one .max file, several glTF scenes: https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#scenes

It's not parenting, it's really a different scene definition; it's much like having separate glTF files.

Right now, the artists have to select the relevant objects and export them to separate gltf files, which takes a large part of their time in some cases.

Selmar:
Actually, come to think of it. If we export several scenes into the same glTF file, we would have to export every scene every time we change one object (which is related to only one scene). This would also cost a lot of time.

I guess a better solution may be to have some kind of scene export configuration; being able to specify which object goes where and whether to reset the pivot, being able to export only the selected "glTF objects", etcetera. Does this make sense to you?

Since we want this kind of tool relatively fast, I would like to implement this this functionality in the babylon exporter. If you are okay with that, we should discuss some implementation details (mostly UI, I guess).

If not, I will probably do something with maxscripts, but this is more difficult to integrate neatly with the exporter.

Selmar:
Perhaps we could even combine the scene and separate gltf export functionalities, but I guess exporting separate gltfs is enough; I will talk to the artists.

Selmar:
I have changed the title to reflect what I would like to add to the exporter.

Instead of having a single file to export to, one would build a list of nodes with export target files. By default it would reference the root node (and thus export everything into one file), with the export target being the name of the `.max` file and the output folder where the `.max` file is.

Some options would be per export, some would be global.

It could be a separate export window if you wish to have both workflows?

Let me know what you prefer.

kcoley:
> Instead of having a single file to export to, one would build a list of nodes with export target files

[...]

PR #380, opened 2018-11-16
Multi-file export (#368)

Selmar:
Implementing #368

This pull request is not necessarily meant to be integrated right away; rather, it is to track and get some feedback on the current state and whether it aligns with your views.

Implemented multi-file export:
- Reset translation of the node used as a scene root.
- Take export settings (file type, etc.) from the exporter window.
- Button to execute multi-file export, with shift-click to modify multi-file export settings: 

image
image

**Issues**:
- Have not tested anything but Max 2016 as I don't have the other versions.
- Have only tested `gltf` format.
- We'll be testing this in-house next week (I will be absent until Thursday 22nov).
- The multi-file export dialog is a modal dialog; this means it blocks interaction with the exporter window. However, this also means it blocks interaction with the 3dsmax window. But, when you click on 3dsmax when the exporter window is not in focus, the exporter window does not show up or become highlighted. It makes it look like 3dsmax is unresponsive. I believe this is due to the exporter window not being a child window of the 3dsmax window (perhaps there was a reason?). We can try changing that, or fix the problem another way.
- Currently, we can not export skins with bones outside of the root node's hierarchy. However, this may be something very useful to have, at least in our case, for example when exporting model LODs that have a shared (animated) hierarchy, or separate characters, etcetera. An alternative (and perhaps better) way to get this to work can be to allow the user to specify layers or selection sets for the file export (in addition to the node?). That way, the different exports can be mixed together in the hierarchy: 
- image


sebavan:
@Selmar , our top exporter guyz are off for the week so the review will happen will happen probably next week. Do not worry if there is a bit of delay :-)

Selmar:
So far this seems to be working well for us. 

Next week I will add a feature to be able to change many file paths at once.

kcoley:
@Selmar nice work!

kcoley:
@Selmar can you add some documentation for this ?
https://github.com/BabylonJS/Documentation

Selmar:
@kcoley I'll see when I can find some time for the documentation, but it won't be very soon.

PR #388, opened 2018-11-22
Guid conflict bugfix mistake

Selmar:
- Guids were not stored until requested the second time; this could potentially cause conflicts.
- Clarify an existing problem with Guid generation in comments (see below). Conflicts may still occur. This must be discussed (will create a separate issue).


sebavan:
Thx adding @deltakosh and @kcoley for review.

PR #416, opened 2018-12-20
Material-ID Allocation fix for Arnold in Max 2019 (and future releases)

Tom-LeFx:
- fixed material-slot allocation for Arnold materials to better withstand changes to slot IDs (due to feature changes on arnold material) -> Only for normal and emissive maps
- also added transfer of emissive color property to glTF export in case a texture exists

- optimized material-alphasettings-export by adding .png to valid extensions when exporting opaque materials


sebavan:
Adding @kcoley to review. Please expect some delays with the vacation period, I guess he will look at it right after the holiday.

Issue #429, opened 2019-01-04
3dsMax: uri with spaces causes glb export to fail


kcoley:
Referenced here:
https://forum.babylonjs.com/t/cant-export-with-1-3-24-version/540/2

kcoley:
Merged a fix

PR #430, opened 2019-01-05
fix glb export

kcoley:
## v1.3.26
**Fixed Bugs:**
- Fixed bug in GLTFImage uri which caused glb files with textures containing spaces to not export
(https://github.com/BabylonJS/Exporters/issues/429)


deltakosh:
This pull request has been mentioned on **Babylon.js**. There might be relevant details there:

https://forum.babylonjs.com/t/cant-export-with-1-3-24-version/540/6

Issue #413, opened 2018-12-18
Max exporter not exporting meshes with multi materials in shell slot


kcoley:
https://forum.babylonjs.com/t/max-exporter-dont-export-meshes-with-multi-materials-in-shell-slot/304

PR #431, opened 2019-01-05
fix shell material multi mat export

kcoley:
## v1.3.27
**Fixed Bugs:**
- Fixed issue where shell material may fail to export
(https://github.com/BabylonJS/Exporters/issues/413)


deltakosh:
This pull request has been mentioned on **Babylon.js**. There might be relevant details there:

https://forum.babylonjs.com/t/max-exporter-dont-export-meshes-with-multi-materials-in-shell-slot/304/9

Issue #436, opened 2019-01-14
"tag" should be "tags" in Max exporter


kcoley:
Original issue here: https://forum.babylonjs.com/t/3dsmax-export-to-gltf-with-tags/686/6

kcoley:
Resolved

PR #438, opened 2019-01-14
added Selmar's multi-select export

kcoley:
## v1.3.30
**Implemented Changes:**
- Added @Selmar 's multi material export
(https://github.com/BabylonJS/Exporters/issues/436)


Selmar:
there's a typo, here and in the change log, it says `multi material export` instead of `multi-file export`

PR #445, opened 2019-01-24
Sync EEVEE branch from master

Palmer-JC:
prepare for moving EEVEE to master


deltakosh:
I guess all that is not related to blender could be overwritten from master

Palmer-JC:
What about 
- making a local copy of EEEVEE version of Blender dir
- switch locally to master & replace Blender dir with EEVEE
- commit & push to master
- Delete EEVEE branch

Issue #454, opened 2019-02-07
Maya: Keys that don't change value in a clip are exported as identity


PatrickRyanMS:
# Animation Issue
There is an issue with the Maya exporter with **node animation** that if the keyed transform value does not change over the duration of the clip when exporting animation groups. *Note this is not an issue with skinned animation as the bones seem to respect keyed TRS values even when they don't change within an animation group.*

For example, you have a node that is keyed to translation 0, 0, 0, on frames 1-10 and keyed to translation 0, 5, 0 on frames 20-30. Those are exported as two animation groups, the first 1-10 and the second 20-30. What will happen is that since the keyed value doesn't change in the clip, a dummy animation is exported and the node is assumed to be at identity. This means the node will be at 0, 0, 0, in both clips rather than passing that the node is actually stationery at 0, 5, 0 in the second clip.

To work around this currently, changing one frame's value at an imperceptible level will force the animation to write out the correct values. So if I keyed the node from the above example to 0, 5, 0 on frames 20-29, and to 0, 5.01, 0 on frame 30 and exported frames 20-30 as an animation group the resulting animation would look right.

The problem with that work around is that keying translation, rotation, scale on many nodes in a file and having to incorporate this work around can be time consuming and error prone. The fix is when exporting the dummy animation clip to export the TRS transform of the node for the clip rather than passing identity.

kcoley:
@PatrickRyanMS just merged a fix for this.  The exporter was determining that an animation group with only two keyframes of the same value were not valid.  I made an update to allow this case to export.

Feierdeifi:
Nice! Thank you very much.

PR #456, opened 2019-02-11
Maya animation group same two key frames

kcoley:
## v1.2.29
**Fixed bugs:**
- Allow animation groups with two key frames of the same value to export


kcoley:
Should resolve: https://github.com/BabylonJS/Exporters/issues/454

PR #480, opened 2019-04-16
Fix older 3dsMax versions compatibility

jbltx:
I have just used the 3dsMax 2017 exporter in *Debug* mode, and it didn't compile due to too recent syntax used for out variables. I have fixed that and also add the possibility to use the exporter in Maxscript without having the dll files in `/bin/assemblies` folder (3dsMax makes a startup check in this folder to initialize plugin with their `Class_ID`).

Here are the different commits :

- Be able to initialize the Babylon Exporter plugin without any Autodesk startup checks.
- Do not use **C# 7.0** syntax for out variables to not break compatibility with older versions of .NET Frameworks used by 3dsMax.
- Use Autodesk standard environment variables instead of guessing drives in VS CSharp projects files.


Drigax:
Thanks for making the binplace path changes in the project files. I've been meaning to clean that up :)

Changes look good. Thanks for contributing!

Issue #466, opened 2019-03-18
[Max][glTF] BaseColor texture turns white when specularLevel too high


Ben-Mack:
In 3ds Max, if material type Standard has high `Specular Level` value (like above 200), the output diffuse texture will become blank white.

This behaviour only occurs on glTF format, texture exported using Babylon format still working fine.

Looking through the source, I think it might related to this function:
https://github.com/BabylonJS/Exporters/blob/5679018c6386043af7ec8c0d4374faeded8c4451/3ds%20Max/Max2Babylon/Exporter/BabylonExporter.GLTFExporter.Material.cs#L589

deltakosh:
Yes the problem is that there is no straightforward transfer between non PBR (standard) and PBR material 

I highly recommend to use PBR directly in 3dsmax if you plan to export to gltf.

Adding @bghgary for additional thoughts

Ben-Mack:
I think this issue should be considered as a bug. Because even when there's no correct Standard to PBR conversion, the exporter should not turns the diffuse texture into blank white...

deltakosh:
WOuld like to get @bghgary thoughts before marking it as a bug

bghgary:
This one is hard for me to judge since I don't know Max's materials. Does exporting to the babylon format and then serializing to glTF have the same issue? If so, can you send me the export babylon file?

Ben-Mack:
I've prepared 2 cases: One with specular level at 200 (named `hi-specular`), one with specular level at 50 (named `lo-specular`)

For each case, I've exported both babylon and gltf format into separate folders.

The material in Max is simply a default Standard material with a bitmap in diffuse slot and the accordingly specular level for each case.

Download: [Blank-tx-on-hi-specular.zip](https://github.com/BabylonJS/Exporters/files/2995795/Blank-tx-on-hi-specular.zip)

bghgary:
I think the Max exporter is still using the PBR spec-gloss conversion to convert a standard material to metal/rough, which is not right. @deltakosh @Drigax

Drigax:
I don't think this is a perfect solution, but reusing our Babylon->GLTF conversion for the Babylon->GLTF converter portion of the 3dsMax exporter should at the very least prevent us from exporting shiny materials as metallic, and improperly changing the material base color based off of its perceived shininess.

PR #485, opened 2019-04-20
Modify Maya/3ds Max GLTF Standard Shader Spec/gloss to PBRMetalRoughness converter

Drigax:
Proposed fix for #466. This modifies the existing Standard Shader specular/gloss to Metal/Roughness converter to use the Babylon engine's specular power -> roughness conversion heuristic.

This change removes the baseColor specular color calculations, and removes the assumptions of surface metallic properties based of its shininess.

### DIFF
--- a/Maya/Exporter/BabylonExporter.GLTFExporter.Material.cs
+++ b/Maya/Exporter/BabylonExporter.GLTFExporter.Material.cs
@@ -571,7 +571,7 @@ namespace Maya2Babylon
             var P3 = new BabylonVector2(1300f, 0.1f);
 
             /**
-             * Helper function that defines the bezier curve as well.Given the control points, solve for x based on a given t for a cubic bezier curve
+             * Helper function that defines the bezier curve as well. Given the control points, solve for x based on a given t for a cubic bezier curve
              * @param t a value between 0 and 1
              * @param p0 first control point
              * @param p1 second control point
--- a/Maya/Exporter/BabylonExporter.GLTFExporter.Material.cs
+++ b/Maya/Exporter/BabylonExporter.GLTFExporter.Material.cs
@@ -564,44 +564,66 @@ namespace Maya2Babylon
 
         private MetallicRoughness ConvertToMetallicRoughness(SpecularGlossiness specularGlossiness, bool displayPrints = false)
         {
-            var diffuse = specularGlossiness.diffuse;
+            // Hard coded points used to define the specular power to roughness curve.
+            var P0 = new BabylonVector2(0f, 1f);
+            var P1 = new BabylonVector2(0f, 0.1f);
+            var P2 = new BabylonVector2(0f, 0.1f);
+            var P3 = new BabylonVector2(1300f, 0.1f);
+
+            /**
+             * Helper function that defines the bezier curve as well.Given the control points, solve for x based on a given t for a cubic bezier curve
+             * @param t a value between 0 and 1
+             * @param p0 first control point
+             * @param p1 second control point
+             * @param p2 third control point
+             * @param p3 fourth control point
+             * @returns number result of cubic bezier curve at the specified t
+             */
+            float _cubicBezierCurve(float t, float p0, float p1, float p2, float p3)
+            {
+                return
+                (
+                    (1 - t) * (1 - t) * (1 - t) * p0 +
+                    3 * (1 - t) * (1 - t) * t * p1 +
+                    3 * (1 - t) * t * t * p2 +
+                    t * t * t * p3
+                );
+            }
+
+            /*
+             * Helper function that calculates a roughness coefficient given a blinn-phong specular power coefficient
+             * @param specularPower the blinn-phong specular power coefficient
+             * @returns number result of specularPower -> roughness conversion curve.
+             */
+            float _solveForRoughness(float specularPower)
+            {
+                var t = Math.Pow(specularPower / P3.X, 0.333333);
+                return _cubicBezierCurve((float)t, P0.Y, P1.Y, P2.Y, P3.Y);
+            }
+
+            var diffuse = specularGlossiness.diffuse.scale(0.5f);
             var opacity = specularGlossiness.opacity;
-            var specular = specularGlossiness.specular;
             var glossiness = specularGlossiness.glossiness;
-
-            var oneMinusSpecularStrength = 1 - specular.getMaxComponent();
-            var metallic = solveMetallic(diffuse.getPerceivedBrightness(), specular.getPerceivedBrightness(), oneMinusSpecularStrength);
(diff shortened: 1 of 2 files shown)


deltakosh:
Pinging @bghgary for review

Maya/Exporter/BabylonExporter.GLTFExporter.Material.cs
```diff
@@ -562,47 +562,69 @@ private bool isTextureOk(BabylonTexture texture)
...
+        private MetallicRoughness ConvertToMetallicRoughness(SpecularGlossiness specularGlossiness, bool displayPrints = false)
+        {
+            // Hard coded points used to define the specular power to roughness curve.
+            var P0 = new BabylonVector2(0f, 1f);
+            var P1 = new BabylonVector2(0f, 0.1f);
+            var P2 = new BabylonVector2(0f, 0.1f);
+            var P3 = new BabylonVector2(1300f, 0.1f);
+
+            /**
+             * Helper function that defines the bezier curve as well.Given the control points, solve for x based on a given t for a cubic bezier curve
```
bghgary:
```suggestion
             * Helper function that defines the bezier curve as well. Given the control points, solve for x based on a given t for a cubic bezier curve
```

Maya/Exporter/BabylonExporter.GLTFExporter.Material.cs
```diff
@@ -562,47 +562,69 @@ private bool isTextureOk(BabylonTexture texture)
...
-            return new MetallicRoughness
-            {
-                baseColor = baseColor,
-                opacity = opacity,
-                metallic = metallic,
-                roughness = 1 - glossiness
-            };
+        private MetallicRoughness ConvertToMetallicRoughness(SpecularGlossiness specularGlossiness, bool displayPrints = false)
+        {
+            // Hard coded points used to define the specular power to roughness curve.
```
bghgary: Are these points supposed to be identical to the glTF serializer in Babylon.js? Are the units for specular power the same between the standard material for Babylon and Max?

Maya/Exporter/BabylonExporter.GLTFExporter.Material.cs
```diff
@@ -562,47 +562,69 @@ private bool isTextureOk(BabylonTexture texture)
...
+            }
+
+            /*
+             * Helper function that calculates a roughness coefficient given a blinn-phong specular power coefficient
+             * @param specularPower the blinn-phong specular power coefficient
+             * @returns number result of specularPower -> roughness conversion curve.
+             */
+            float _solveForRoughness(float specularPower)
+            {
+                var t = Math.Pow(specularPower / P3.X, 0.333333);
```
bghgary: I can see this is copied from Babylon.js, but I can't remember why there is a power to the 1/3 here. @kcoley Do you remember?

Maya/Exporter/BabylonExporter.GLTFExporter.Material.cs
```diff
@@ -562,47 +562,69 @@ private bool isTextureOk(BabylonTexture texture)
...
+                return _cubicBezierCurve((float)t, P0.Y, P1.Y, P2.Y, P3.Y);
+            }
+
+            var diffuse = specularGlossiness.diffuse.scale(0.5f);
+            var opacity = specularGlossiness.opacity;
+            var glossiness = specularGlossiness.glossiness;
+            var metallic = 0;
+            var roughness = _solveForRoughness(glossiness * 256); // Glossiness = specularPower / 256
```
bghgary: Where does 256 come from?

Drigax: When we construct our SpecularGlossiness structure from the babylonMaterial we converted, we divide the specularPower by 256. I'm not entirely sure why...

https://github.com/BabylonJS/Exporters/blob/b45ec95831d630faff22f291192d4be542a11330/3ds%20Max/Max2Babylon/Exporter/BabylonExporter.GLTFExporter.Material.cs#L186

bghgary: But this is not a Babylon material right?

Drigax: and when we initially convert from Max->Babylon we, we multiply the shininess (Gloss in Max) by 256 to become what we define as SpecularPower in Babylon...strange...

Looking at the source, the shininess of a material is represented in max as a 0-1 value, while we represent it in Babylon as a 0-infinity value. I'm not sure why we use a factor of 256 other than to get this "looking right".

https://github.com/BabylonJS/Exporters/blob/b45ec95831d630faff22f291192d4be542a11330/3ds%20Max/Max2Babylon/Exporter/BabylonExporter.Material.cs#L152

Maya/Exporter/BabylonExporter.GLTFExporter.Material.cs
Drigax: yes, but not really?

here's two test renders I made using a directional light  and spheres of varying SpecularPower(Babylon)/glossiness(3dsMax) using the same values:

Babylon:
specspectum_babylon (3)

3dsMax:
specspectum_3dsmax

Maya/Exporter/BabylonExporter.GLTFExporter.Material.cs
Drigax: If we plot out our converter curve, we get this nice hyperbolic-ish curve:
shininess to roughness curve

Maya/Exporter/BabylonExporter.GLTFExporter.Material.cs
Drigax: It gets even worse when exporting from Max to Babylon, we get this:
screenshot_19-4-22_17-22


I'm willing to accept the disparity between different implementaitons of the Blinn lighting model as an unsolved mystery.

Drigax: as we discussed offline, this is caused by the 3dsMax->Babylon portion of the converter, I'm not quite sure if there is a linear relationship between 3dsMax and BabylonJS's shininess units.

But, I believe for the sake of "correctness" we can assume in our Babylon->GLTF part of the conversion, we are only working in Babylon units.

Maya/Exporter/BabylonExporter.GLTFExporter.Material.cs
Drigax: I'll most likely do a deeper study in whether or not this is correct when I get around to tackling the 3dsMax->Babylon standard material conversion as well.

Hey Kacey, if you still have access to the data that defined this original curve, that would be a huge help.

PR #491, opened 2019-04-24
Custom material exporters

elpie89:
I repropose the custom Material exporter, it should be possible to use a custom material with gltf
on fdb073d I changed also the way the check on other material was made.
I use ClassID instead of names, there was a pending todo on this part.
Please this brunch because it is very important for many studios (like mine) to work with custom material


Drigax:
There's alot of changes in this pull request not related to Custom materials like your changes to AnimationGroup. Can you minimize this pull request to only include the Custom Material-related changes?

3ds Max/Max2Babylon/Exporter/BabylonExporter.GLTFExporter.Texture.cs
```diff
@@ -107,31 +108,31 @@ private GLTFTextureInfo ExportTexture(BabylonTexture babylonTexture, GLTF gltf,
...
-                // Mag and min filters
-                GLTFSampler.TextureMagFilter? magFilter;
-                GLTFSampler.TextureMinFilter? minFilter;
-                getSamplingParameters(babylonTexture.samplingMode, out magFilter, out minFilter);
-                gltfSampler.magFilter = magFilter;
-                gltfSampler.minFilter = minFilter;
-                // WrapS and wrapT
-                gltfSampler.wrapS = getWrapMode(babylonTexture.wrapU);
-                gltfSampler.wrapT = getWrapMode(babylonTexture.wrapV);
+            // Mag and min filters
```
Drigax: Can you please not change the indentation?

3ds Max/Max2Babylon/Exporter/BabylonExporter.GLTFExporter.Texture.cs
```diff
@@ -144,11 +145,11 @@ private GLTFTextureInfo ExportTexture(BabylonTexture babylonTexture, GLTF gltf,
                 }
 
 
-                // --------------------------
-                // --------- Image ----------
-                // --------------------------
+            // --------------------------
```
Drigax: Indentation is messed up here too

3ds Max/Max2Babylon/Exporter/BabylonExporter.GLTFExporter.Texture.cs
```diff
@@ -174,60 +175,60 @@ private GLTFTextureInfo ExportTexture(BabylonTexture babylonTexture, GLTF gltf,
...
-
-                // --------------------------
-                // ------ TextureInfo -------
-                // --------------------------
-                var gltfTextureInfo = new GLTFTextureInfo
-                {
-                    index = gltfTexture.index,
-                    texCoord = babylonTexture.coordinatesIndex
-                };
+            // --------------------------
```
Drigax: Indentation here too

3ds Max/Max2Babylon/Exporter/BabylonExporter.GLTFExporter.Texture.cs
```diff
@@ -174,60 +175,60 @@ private GLTFTextureInfo ExportTexture(BabylonTexture babylonTexture, GLTF gltf,
...
-                    }
                 }
+            }
                 var textureID = name + TextureTransformID(gltfTextureInfo);
                 // Check for texture optimization.  This is done here after the texture transform has been potentially applied to the texture extension
                 if (CheckIfImageIsRegistered(textureID))
                 {
                     var textureComponent = GetRegisteredTexture(textureID);
-
+
```
Drigax: whitespace changes here as well, please use a separate PR to correct whitespace where there's no logic change.

3ds Max/Max2Babylon/Exporter/BabylonExporter.IMaterialExporter.cs
```diff
@@ -0,0 +1,136 @@
...
+    {
+        ClassIDWrapper MaterialClassID { get; }
+    }
+
+    public interface IBabylonMaterialExporter : IMaterialExporter
+    {
+        BabylonMaterial ExportBabylonMaterial(IIGameMaterial material);
+    }
+
+    delegate string TryWriteImageCallback(string sourceTexturePath);
```
Drigax: Do we ever use this delegate?

Drigax:
I'm also somewhat hesitant to add features to our exporters that aren't clearly defined as part of the Babylon or GLTF spec. This may work well for your use, but I'd like to know more about usecases and where this functionality is applicable so we don't add functionality that nobody uses

3ds Max/Max2Babylon/Exporter/AnimationGroup.cs
```diff
@@ -162,6 +198,19 @@ public void LoadFromData(string propertyName)
...
             }
 
+            AnimationGroupNodes = new List<AnimationGroupNode>();
+            foreach (uint nodeHandle in nodeHandles)
+            {
+                var node = Loader.Core.GetINodeByHandle(nodeHandle);
+                if (node != null)
+                {
+                    string name = node.Name;
+                    string parentName = Loader.Core.GetINodeByHandle(nodeHandle).ParentNode.Name;
```
Drigax: This looks like the same code in your other PR

3ds Max/Max2Babylon/Exporter/BabylonExporter.GLTFExporter.Texture.cs
```diff
@@ -58,10 +58,12 @@ private GLTFTextureInfo ExportBitmapTexture(GLTF gltf, BabylonTexture babylonTex
 
         private GLTFTextureInfo ExportTexture(BabylonTexture babylonTexture, GLTF gltf)
         {
-            return ExportTexture(babylonTexture, gltf, null, () =>
-            {
-                var sourcePath = babylonTexture.originalPath;
+            return ExportTexture(babylonTexture, gltf, null, 
+                () => { return TryWriteImage(gltf, babylonTexture.originalPath, babylonTexture.name); });
```
Drigax: Why do we want to refactor this into another method?

3ds Max/Max2Babylon/Exporter/BabylonExporter.IMaterialExporter.cs
```diff
@@ -0,0 +1,136 @@
...
+        GLTFMaterial ExportGLTFMaterial(BabylonExporter exporter, GLTF gltf, IIGameMaterial material, Func<string, string, string> tryWriteImageFunc,
+            Action<string, Color> raiseMessageAction, Action<string> raiseWarningAction, Action<string> raiseErrorAction);
+    }
+
+    // We require a separate struct, because the IClass_ID does not implement GetHashCode etc. to work with dictionaries
+    public struct ClassIDWrapper : IEquatable<ClassIDWrapper>
+    {
+        public static ClassIDWrapper XRef_Material = new ClassIDWrapper(0x272c0d4b, 0x432a414b);
+        public static ClassIDWrapper Advanced_Lighting_Override_Material = new ClassIDWrapper(0x2914493d, 0x6cff42f7);
+        public static ClassIDWrapper Morpher_Material = new ClassIDWrapper(0x4b9937e0, 0x3a1c3da4);
```
Drigax: Can you explain where these class ID hex values are received from? Also, It may be best to define them as constants, for the sake of reducing our magic numbers

Drigax: Also, how confident are you that these won't change with an update to Max?

elpie89:
I understand I will check again with Selmar to make this pull request easier to integrate.
I will spend some time as soon as possible to clarify some points.
Can you take a look at the other pull request? Should be clear and easier to test. Thanks for your time

3ds Max/Max2Babylon/Exporter/BabylonExporter.IMaterialExporter.cs
elpie89: Yes we can define this constant.
About the source.
I executed this simple pyhton script in 3dsmax

```
import MaxPlus

for cd in MaxPlus.PluginManager.GetClassList().Classes:
    if cd.SuperClassId == MaxPlus.SuperClassIds.Material:
        anim = MaxPlus.Factory.CreateAnimatable(MaxPlus.SuperClassIds.Material, cd.ClassId, False)
        if anim:
            inst =  MaxPlus.Mtl._CastFrom(anim)
            if inst:
                print str(inst) + " has: " + str( cd.ClassId)
```
About the confidence:
As the class ID are generated randomly and manually I do not think Autodesk will never change any of them. This will potentially cause a lot of trouble for many third part developer

3ds Max/Max2Babylon/Exporter/BabylonExporter.IMaterialExporter.cs
elpie89: this is not used inside babylon but we use it to export image from our custom material exporter

3ds Max/Max2Babylon/Exporter/BabylonExporter.GLTFExporter.Texture.cs
elpie89: see above, like for the delegate

Drigax:
Closing due to #493

3ds Max/Max2Babylon/Exporter/BabylonExporter.IMaterialExporter.cs
Drigax: Ok, I think going with constants is the right way to handle something like this. Mainly because if 3DS Max 2025 or something changes the ID's then for our exporter all we'd need to do is to use #defines for the exporter version

PR #492, opened 2019-04-24
Animation group export

elpie89:
Our artists need the possibility to export and import animation group
Sometimes they work on a file with 100 animation group. then for some reason, they have to move all this info into a new file
This allows to do it


elpie89:
Can we proceed with this pull request maybe before the other one?

Drigax:
Sure, can you help me understand why we would want to integrate this into the project? I can see that we're adding extra serialization for animation groups, and exporting to a separate file, but why would an end user want this?

Can you help explain your company's pipeline, and how they use the animation group list, and why exporting as part of the Babylon file isn't sufficient?

elpie89:
Of course.
Practically working in a production environment many studios can work on the same object for months.
When I say object I mean something that at the end of the pipeline will be only one GLTF
The outsourcer, for instance, can work on a different part of a car(for example), and a manager is responsible to grab all modification made from that outsourcer and merge this in a new max file.
Now the production goes on the max file and all animation group are done, but then polish on a part of the car is requested.
The manager most of the time get from the outsourcer only the mesh updated not the animation group. 
And many hours are spent recreating animation group manually
It is easy to export an fbx and merge it into a new file, but the animation group are stored in the hide scene root of the max file. 
Another example is that when you spend hours working on animation groups for a lod0, then should be easy to export this info into a Lod1 and so on.

elpie89:
> Do you have a custom solution for interpreting the created AnimationGroup file as well?

About this, no actually, I mean we have a custom engine, but the animation group are interpreted as with a normal GLTF viewer

Drigax:
Ah, I see! Thanks for elaborating. 

@bghgary and @PatrickRyanMS, I'm interested in getting your perspectives as well. I'm starting to understand the utility of adding hooks into our exporters to support this extension of functionality, I'm mostly just concerned about generating .json files that aren't intended to be consumed by our importing engine, and that we're adding another serialization path just for animation groups.

Am I being overly pedantic on this?

PatrickRyanMS:
This approach is actually very common in games where you have multiple assets feeding one final character on screen. One asset has a base mesh with the default rig and some or all animations. Another rig has the final mesh with textures and default rig. And there are instances where each animation clip is a separate asset. In the final engine, you path animations from the source assets to the final mesh asset so in engine it appears that the final mesh has all of the animations even though they are being pathed from a different (or multiple different) sources. 

The benefit this has from a creation standpoint is that you can have multiple creators working on their part of the asset all at the same time. Modeling and texturing can be done by different people, and animation clips can each be assigned to different people. As long as the rig is common between all assets, you can be editing every part of the asset package at once and not stomping changes or losing time. 

For glTF, this is a problem because we need to write all of the animations into the glTF file to work in a viewer as a stand alone asset and don't have the benefit of an engine. Even if we were pulling this into Babylon, we could push the animations between assets. But for a single file, our current pipeline means you need to place all animation into the file you are exporting. I see a path to where we may want to have an option to pull animation groups from another (or several other) Max/Maya files when writing out the glTF. Now does that tool live in the Max/Maya exporter or would that be better served by a tool in the sandbox, for example, where someone using Blender or some other DCC tool could also take advantage of it. 

Taking it out of our exporter makes it available to more tools that we aren't building exporters for, but at the same time, we add another step in the pipeline. Seems to me that the best of all worlds would be to have a system like this in the exporter combining meshes and animations from Max/Maya files and have an extension to the sandbox where we could load a file, and have a UI loader for animation pulled from other glTF/OBJ files that we could save out to a glTF with the current save functionality in the inspector. Obviously, we are missing FBX from that scenario, but that isn't currently supported in the sandbox.

Drigax:
After talking with @bghgary and @PatrickRyanMS offline, I'm convinced that this may be a workaround for a bigger problem, which is how can one assemble assets from multiple 3d scenes to export as a single GLTF. Currently the GLTF spec doesn't really support this well, we can't trivially retarget animations exported with a skinned mesh onto another (like say retargeting the same animations for multiple LOD meshes)

Right now, I'm not sure that shimming the exporter is the right way to go. I think adding the additional modularity for Animation Groups and hooks to support your export pipeline is a good start, but I don't think that the JSON export is something I want to integrate mainline right now.

Thanks for the contribution!

Selmar:
Alright. I understand your viewpoint. It's a bit of a hassle for us to manage several branches (mostly when conflicting changes happen), but nothing we can't live with.

deltakosh:
@Drigax Maybe we can offer @selmar a ETA for a better solution?

Drigax:
I doubt I can get to it in a reasonable amount of time, since this would be architecting something from scratch. I'll add the current implementation to support your workflow, and refactor it later on when we have something better.

PR #494, opened 2019-04-26
fix skin node hierarchy

### LINKED ISSUE #489: Exporting skinned data from max use wrong bones
I attached a sample,
Basically, a mesh that is skinned to 2 bones is actually using the entire hierarchy from the root to this 2 bones

```
 "skins": [
        {
            "inverseBindMatrices": 6,
            "skeleton": 0,
            "joints": [
                0, // this shoudl not be here
                1,
                2
            ],
            "name": "skeleton #0"
        }
    ]
```

This is part of the sample I attached  as you see in the data 3 bones are referenced, looking on the code this is because the way "relative nodes " are calculated

In other when a vertex is not weighted to any bones it automatically fallback in the root of his hierarchy.
Should be notified to the artist and is weight should fallback on the first skinned bone, 
not on the root of the hierarchy
[BabylonTest.zip](https://github.com/BabylonJS/Exporters/files/3111593/BabylonTest.zip)

elpie89:
Fixes #489 
@Selmar 
Exporting bones of a mesh there is no need to export the entire hierarchy with them

### DIFF
--- a/3ds Max/Max2Babylon/Exporter/BabylonExporter.Skeleton.cs
+++ b/3ds Max/Max2Babylon/Exporter/BabylonExporter.Skeleton.cs
@@ -61,7 +61,6 @@ namespace Max2Babylon
                 return revelantNodesBySkin[skin];
             }
 
-            List<IIGameNode> revelantNodes = new List<IIGameNode>();
             List<IIGameNode> bones = GetBones(skin);
 
             if(bones.Contains(null))
@@ -70,31 +69,9 @@ namespace Max2Babylon
                 RaiseError("The skin cannot be exported", logRank);
                 return new List<IIGameNode>();
             }
-            
-            // for each bone of the skin, add their parents in the revelantNodes list.
-            foreach (IIGameNode bone in bones)
-            {
-                if (!revelantNodes.Contains(bone))
-                {
-                    revelantNodes.Add(bone);
-                    IIGameNode currentNode = bone.NodeParent;
-                    while (currentNode != null)
-                    {
-                        if (!revelantNodes.Contains(currentNode))
-                        {
-                            revelantNodes.Add(currentNode);
-                            currentNode = currentNode.NodeParent;
-                        }
-                        else // the node and its parents are already in the list
-                        {
-                            currentNode = null;
-                        }
-                    }
-                }
-            }
 
-            // return an empty list if there is more than one root
-            List<IIGameNode> rootNodes = revelantNodes.FindAll(node => node.NodeParent == null);    // should contains only one node
+            //return a list of all bones that are root
+            List<IIGameNode> rootNodes = GetSkinRoots(bones,skin);
             if (rootNodes.Count > 1)
             {
                 string rootNames = "";
@@ -121,7 +98,7 @@ namespace Max2Babylon
             {
                 IIGameNode currentNode = siblings.Pop();
 
-                if (revelantNodes.Contains(currentNode))    // The node is part of the skeleton
+                if (bones.Contains(currentNode))    // The node is part of the skeleton
                 {
                     // Add the current node to the sorted list
                     sorted.Add(currentNode);
@@ -135,10 +112,25 @@ namespace Max2Babylon
                 }
             }
 
-            revelantNodes = sorted;
-            revelantNodesBySkin.Add(skin, revelantNodes);   // Stock the result for optimization
+            revelantNodesBySkin.Add(skin, sorted);   // Stock the result for optimization
 
-            return revelantNodes;
+            return sorted;
+        }
+
+        private List<IIGameNode> GetSkinRoots(List<IIGameNode> relevantNodes, IIGameSkin skin)
+        {
+            List<IIGameNode> relevantRootNodes = new List<IIGameNode>();
+            List<IIGameNode> skinnedNodes = GetBones(skin);
+            foreach (IIGameNode relevantNode in relevantNodes)
+            {
+                IIGameNode parent = relevantNode.NodeParent;
+                if (!skinnedNodes.Contains(parent))
+                {
+                    //found a skin bone root
+                    relevantRootNodes.Add(relevantNode);
+                }
+            }
+            return relevantRootNodes;
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


3ds Max/Max2Babylon/Exporter/BabylonExporter.Skeleton.cs
```diff
@@ -61,67 +61,42 @@ private List<IIGameNode> GetRevelantNodes(IIGameSkin skin)
...
+            //return a list of all bones that are root
+            IIGameNode rootNodes = GetCommonAncestor(bones);
+            if (rootNodes == null)
             {
                 string rootNames = "";
-                foreach(IIGameNode root in rootNodes)
-                {
-                    rootNames += $" {root.Name}";
-                }
                 RaiseError($"More than one root node for the skin: {rootNames}", logRank);
```
Drigax: Can we update the debug outputs and function docstring to reflect the change in logic?

3ds Max/Max2Babylon/Exporter/BabylonExporter.Skeleton.cs
```diff
@@ -61,67 +61,42 @@ private List<IIGameNode> GetRevelantNodes(IIGameSkin skin)
...
+                RaiseError("Skin has bones that are outside of the exported hierarchy.", logRank);
+                RaiseError("The skin cannot be exported", logRank);
+                return new List<IIGameNode>();
             }
 
-            // return an empty list if there is more than one root
-            List<IIGameNode> rootNodes = revelantNodes.FindAll(node => node.NodeParent == null);    // should contains only one node
-            if (rootNodes.Count > 1)
+            //return a list of all bones that are root
+            IIGameNode rootNodes = GetCommonAncestor(bones);
```
Drigax: NIT: also now we only have a single rootNode

Drigax:
Thanks for the quick fix! can you update the pull request text to be more descriptive of the issue observed, and what change was made? This will make reviewing code much easier and help us keep a clean change history.

Also, can you provide a test scene as well?

3ds Max/Max2Babylon/Exporter/BabylonExporter.Skeleton.cs
```diff
@@ -135,10 +110,88 @@ private List<IIGameNode> GetRevelantNodes(IIGameSkin skin)
...
+
+            List<IIGameNode> previousANodes = new List<IIGameNode>();
+            previousANodes.Add(nodeA);
+            List<IIGameNode> previousBNodes = new List<IIGameNode>();
+            previousBNodes.Add(nodeB);
+
+            int previousANodesIndex = 0;
+            int previousBNodesIndex = 0;
+
+            while (true)
```
Drigax: so here, we populate "previousANodes" and "previousBNodes" with the hierarchy of A and B. Can we change the name to be a bit more descriptive?

3ds Max/Max2Babylon/Exporter/BabylonExporter.Skeleton.cs
```diff
@@ -135,10 +110,88 @@ private List<IIGameNode> GetRevelantNodes(IIGameSkin skin)
...
+                {
+                    break;
+                }
+
+
+            }
+
+            IIGameNode topA = previousANodes[previousANodesIndex--];
+            IIGameNode topB = previousBNodes[previousBNodesIndex--];
+            if (topA.MaxNode.Handle != topB.MaxNode.Handle)
```
Drigax: Here we check to make sure that nodes A and B exist in the same hierarchy?

3ds Max/Max2Babylon/Exporter/BabylonExporter.Skeleton.cs
```diff
@@ -135,10 +110,88 @@ private List<IIGameNode> GetRevelantNodes(IIGameSkin skin)
...
+            {
+                return null;
+            }
+
+            
+
+            while (true)
+            {
+                // in the case that one is a child of the other, other will be invalid when one still has more nodes to go
```
Drigax: Here's where the magic happens, and we walk the hierarchy to find the shared ancestor.

if we have a tree like:
image

And our inputs are D and F, do we expect to return A or C?
Given the function name "GetCommonAncestor", I'd assume that we want C

Likewise, for a case of F and C, our expected output should be C as well, right?

Can we create a scene to test this functionality?

Selmar: Yes, C. In fact in the original code the name is GetLowestCommonAncestor. It is a problem in our engine for the skeleton hierarchy to contain the entire chain up to A, because of culling. Our `skin` object is attached to the node just above the skeleton root. If B was animated to move a long distance, and the skeleton root is on or above A, the bounding box of the skin would not move with the animation of B. This results in the object disappearing when you do not look at the t-pose position of the bounding box, even though there is not even any skinning or skinned mesh on or above B.

3ds Max/Max2Babylon/Exporter/BabylonExporter.Skeleton.cs
Selmar: Yes.

3ds Max/Max2Babylon/Exporter/BabylonExporter.Skeleton.cs
```diff
@@ -61,67 +61,41 @@ private List<IIGameNode> GetRevelantNodes(IIGameSkin skin)
...
-            if (rootNodes.Count > 1)
+            IIGameNode sharedRootNode = GetLowestCommonAncestor(bones);
+            if (sharedRootNode == null)
             {
                 string rootNames = "";
-                foreach(IIGameNode root in rootNodes)
-                {
-                    rootNames += $" {root.Name}";
-                }
                 RaiseError($"More than one root node for the skin: {rootNames}", logRank);
```
Drigax: Line 83's debug statement still doesn't seem correct

Selmar: Correct. There is also another problem, because the node hierarchy exported is currently not correct when there are gaps in the bone hierarchy.

I'm fixing this right now, but we will not test it before monday. @elpie89

elpie89: I made some tests, and everything works on my side, I tested with many different hierarchies

Drigax: Great, can you please upload the test scenes as part of this conversation? I'll approve the PR

elpie89:
Hi @Drigax, thanks for accepting the PR
This is the file you requested 
[HierarchyTest.zip](https://github.com/BabylonJS/Exporters/files/3130707/HierarchyTest.zip)

PR #500, opened 2019-05-02
fix nullref on standard physical material export

elpie89:
I had some problem with line endings, but the issue is fixed now
To test the problem before the fix just :
create a sphere assign a Standard Physical Material without texture
export


deltakosh:
pinging @drigax

Drigax:
Why is all the whitespace changed again? this is making review really difficult.

are you developing on windows and not checking in with Unix style line endings?

PR #510, opened 2019-05-15
Implement 3DS Max double-sided materials, and some minor build fixes.

jdwilder-google:
This change adds support for double-sided materials in 3DS Max via the 'Double Sided' material. It also includes some minor fixes for build errors I encountered compiling in VS 2015.

PR #513, opened 2019-05-17
VS 2015 build fixes for 3DS Max Exporter

jdwilder-google:
Specifically:
- Local functions are not supported, so this extracts them to the class level.
- Declaration in function call parameter is not supported, so this moves them out to the containing scope.
- Skip copying DLLs for any versions of 3DS Max that are not installed, so copy errors don't cause a build error.


jdwilder-google:
Split from #510

PR #515, opened 2019-05-20
Relative file path

elpie89:
force ExportItem class to keep consistency with multiexport 
path of gltf is now stored relative to the project folder if project folder exists

### DIFF
--- a/3ds Max/Max2Babylon/Tools/Tools.cs
+++ b/3ds Max/Max2Babylon/Tools/Tools.cs
@@ -56,7 +56,7 @@ namespace Max2Babylon
         /**
          * Computes a texture transform matrix with a pre-transformation
          */
-        public static BabylonMatrix ComputeTextureTransformMatrix(BabylonVector3 pivotCenter, BabylonVector3 offset, BabylonQuaternion rotation, BabylonVector3 scale)
+        public static BabylonMatrix ComputeTextureTransformMatrix(BabylonVector3 pivotCenter , BabylonVector3 offset, BabylonQuaternion rotation, BabylonVector3 scale)
         {
             var dOffset = new BabylonVector3();
             var dRotation = new BabylonQuaternion();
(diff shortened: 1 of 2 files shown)


3ds Max/Max2Babylon/Tools/Tools.cs
```diff
@@ -1130,6 +1131,70 @@ public static string GetRelativePath(string fromPath, string toPath)
             return relativePath;
         }
 
+
+        public static string UnformatPath(string formattedPath)
+        {
+            string newPath = formattedPath;
+            newPath = Regex.Replace(formattedPath, @"[()]", string.Empty);
```
Drigax: I'm not entirely sure what your usecase is here.

I was assuming that this allows us to use relative pathing shortcuts like ./ ../ and ~/ when indicating the output directory for the babylon exporter, and we would be able to resolve that to the full path on export but this doesn't appear to do that.

Can you please give me some context behind these changes and why they're useful? It would be especially great if I can get some testing scenarios.

deltakosh:
Up @elpie89 :)

3ds Max/Max2Babylon/Tools/Tools.cs
elpie89: My idea is that we should try, when possible, to store the path relative to the project folder,
but at the same time always display the absolute bath on the user who grabs the file
That is the reason why I added the format/unformat methods to display the project folder part of the path around parenthesis
something like this
user choice:
c:\my_max_project\assets\3d\gltf\model.gltf
we display
(c:\my_max_project)\assets\3d\gltf\model.gltf
we store
\assets\3d\gltf\model.gltf

user B open the max file with the project folder configured to other location
we display
(d:\3ds_max_projects\my_max_project)\assets\3d\gltf\model.gltf

Drigax:
I see! That's a very good usecase, I definitely run into this quite often when opening other user's POC's. Thanks for the fix.

PR #516, opened 2019-05-20
Get animation group by node

### LINKED ISSUE #228: [Max][AnimationGroups] Find AnimationGroup(s) that use(s) a node
Artists requested to have a way to select the animation groups in which a max node is used. Because we have a lot of animations (>100 in some cases), the artists could really use this to fix things (e.g. to figure out if a mesh is used in the wrong animation).

elpie89:
Fixes #228

### DIFF
--- a/3ds Max/Max2Babylon/Forms/AnimationForm.cs
+++ b/3ds Max/Max2Babylon/Forms/AnimationForm.cs
@@ -45,7 +45,7 @@ namespace Max2Babylon
         private void createAnimationButton_Click(object sender, EventArgs e)
         {
             AnimationGroup info = new AnimationGroup();
-
+            
             // get a unique name and guid
             string baseName = info.Name;
             int i = 0;
@@ -88,7 +88,7 @@ namespace Max2Babylon
                 return;
 
             AnimationGroup selectedItem = (AnimationGroup)AnimationListBox.SelectedItem;
-
+            
             // delete item
             selectedItem.DeleteFromData();
 
--- a/3ds Max/Max2Babylon/BabylonAnimationActionItem.cs
+++ b/3ds Max/Max2Babylon/BabylonAnimationActionItem.cs
@@ -28,7 +28,6 @@ namespace Max2Babylon
 
             form.HighlightAnimationGroupOfSelection();
 
-
             return true;
 		}
 


deltakosh:
Thanks a lot for your appreciated help!!

pinging @Drigax

3ds Max/Max2Babylon/Forms/AnimationForm.cs
```diff
@@ -45,7 +45,7 @@ private void AnimationForm_Load(object sender, EventArgs e)
         private void createAnimationButton_Click(object sender, EventArgs e)
         {
             AnimationGroup info = new AnimationGroup();
-            
+
```
Drigax: can we please minimize white space changes not related to code changes. it makes git-blame associate you with stuff you may not have changed substantially.

I always appreciate code cleanup, but I'd like to avoid hygienic changes that aren't part of a behavioral change.

3ds Max/Max2Babylon/Forms/AnimationForm.cs
```diff
@@ -88,7 +88,7 @@ private void deleteAnimationButton_Click(object sender, EventArgs e)
                 return;
 
             AnimationGroup selectedItem = (AnimationGroup)AnimationListBox.SelectedItem;
```
Drigax: same here, we can do this in a separate PR

deltakosh:
Up @elpie89

elpie89:
Should be fixed now, this white space stuff is getting me crazy,
I'm trying to understand how to configure vs to avoid this.
Do we have any guidelines for developers who join the project?

Drigax:
Best practice is to usually look over your changeset before commit via `git diff`. I'm assuming you're developing on windows as well, best thing to do in this case is to configure git to checkout windows-style, but commit unix-style line endings, so that git doesn't mangle them for a whole file on checkin.

I usually enable visual whitespace in VS, so that I can see spaces as `.`

Also, it never hurts to look over your pull request after creation

elpie89:
Yes, you are right, I'm pretty new to open source project, and I always worked on perforce.
I'll try to avoid these kinds of problems in the future.

Drigax:
Thanks for the contribution! changes look good!

Drigax:
I broke the project when resolving the merge conflicts from this commit. I'll have to retry merging this manually.

elpie89:
@Drigax 
this merge broke something, multiselection, clean and merge are no longer there
Let me know if I can help from my side

Drigax:
hey @elpie89 yeah, I had to revert the original merge commit. I re-added them via #527. can you check that this properly re-implemented them before I publish the next build number?

PR #524, opened 2019-05-29
Update mesh.py

zebus3d:
[ FIXBUG ] AtrributeError Context object has no attribute depsgraph


Blender/src/babylon_js/mesh.py
```diff
@@ -167,7 +167,11 @@ def __init__(self, bpyMesh, scene, exporter):
                 Logger.warn('No materials have been assigned: ', 2)
 
         # Get mesh temporary version of mesh with modifiers applied
-        mesh = bpyMesh.to_mesh(bpy.context.depsgraph, True)
+        if bpyMesh.type == 'MESH':
+            # mesh = bpyMesh.to_mesh(bpy.context.depsgraph, True) # old
```
sjaustirni: Maybe it is not necessary to have this `# old `line here?

zebus3d: No, you don't need that commented line, you can erase it.

sjaustirni: Would you update your branch with a commit that deletes this line?

zebus3d:
> This is essentially what I did. I had already made changes for version 6.1 on the master branch when this happened, & did not want to have to put out a patch.
> 
> Please also change the version number in top level **init**.py to (6, 0, -13). FYI, -1 is alpha -2 is beta 1, so -13 is beta 12.
> 
> Do you also have to zip?

I have already updated the Blender2Babylon-6.0.zip file, and also the init to version 6, 0, -13 as you asked me.

Palmer-JC:
Thanks, for the patches.  6.1 with env files, additional the PBR properties anistropy, sub-surface tint, clear coat, & sheen is out in a day or 2.  6.0's days are numbered.

Had not see the Materials Panel error before. Nice catch.

PR #526, opened 2019-05-31
Development

zebus3d:


Palmer-JC:
I have already done most of these for 6.1.  As Blender will be moving to its own repo, I see no reason to continue to do PR here.  Will look at all files here to ensure done.  ETA, sometime this week.

PR #528, opened 2019-06-04
fix merge, clean button after last merge issues

elpie89:
some buttons in the designer were missing.
a control was causing a null reference exception


elpie89:
to avoid this kind of problems, in the future, if you scheduled to do a UI/UX refactoring, I suggest using WPF instead of WinForms

Drigax:
Thanks for the fix, I'll merge this in order with the next PRs so we avoid conflicts

PR #530, opened 2019-06-04
3ds Max exporter batching improvements, add parameter saving, decouple export entrypoint from export button click callback.

elpie89:
InitParameters() now get all saved options inside max
added "Save" button to force options to be registered
added export function with no parameters to match Babylon

basically, the idea is to have the possibility to batch exporter operation
to do this we could call Export() from max script, but there was no match between Babylon configuration and the export function,
in other, it was the internal export button that registered user options,
passing from the max script means that user need the possibility to force to save the specified options

I also exposed the merge and import  animation group from a JSON file
There are many tools that need to be built around this exporter.
I'm planning to expose all the animation group methods soon

### DIFF
--- a/3ds Max/Max2Babylon/ExportParameters.cs
+++ b/3ds Max/Max2Babylon/ExportParameters.cs
@@ -21,5 +21,9 @@
         public bool enableKHRTextureTransform = false;
         public bool enableKHRMaterialsUnlit = false;
         public Autodesk.Max.IINode exportNode;
+
+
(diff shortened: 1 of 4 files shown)


elpie89:
I had to merge the "texture folder explicit" branch into this one
There were too many dependencies
They should just be evaluated in the correct order
texture_folder_explicit before
than this one

sebavan:
@Drigax could you double check this so that we can merge it before the UI work ?

PR #531, opened 2019-06-04
Python script for Changelog auto-completion

Jaskar:
Python script for Changelog auto-completion, retrieving datas from the Github.

- Changelog : Last version date added (needed for auto completion)
- Readme : Explaination added
- BabylonExporter.cs : Version number incremented by 0.1 for testing purpose

+ Merge from official repo in mine

### DIFF
--- a/3ds Max/Max2Babylon/Max2Babylon_Package.py
+++ b/3ds Max/Max2Babylon/Max2Babylon_Package.py
@@ -22,7 +22,7 @@ lastVersionTime = ''
 # Retrieved for the "Exporter/BabylonExporter.cs" file
 
 Max2BabylonVersion = ''
-with open('Max2Babylon/Exporter/BabylonExporter.cs', 'r') as file_BabylonExporterCS:
+with open('Exporter/BabylonExporter.cs', 'r') as file_BabylonExporterCS:
     lineIndex = 0
     for line in file_BabylonExporterCS:
         lineIndex += 1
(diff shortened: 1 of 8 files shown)


3ds Max/Max2Babylon/Max2Babylon_Package.py
```diff
@@ -1,14 +1,137 @@
...
+    }
+    """
+    query = {'query': queryString.replace('\n', ' ')}
+    headers = {'Authorization': 'token ' + accessToken,
+            'Content-Type': 'application/json'}
+
+    response = requests.post(
+        'https://api.github.com/graphql', headers=headers, json=query)
+
+    if (json.loads(response.text)['message']) :
```
Drigax: If the response is data, then it wouldn't include a message, so this would cause us get a KeyError.

3ds Max/Max2Babylon/Max2Babylon_Package.py
```diff
@@ -1,14 +1,137 @@
...
 import shutil
+import json     # Changelog
+import requests  # Changelog
+import datetime  # Changelog
 
 versions = ['2015', '2017', '2018', '2019']
 projectFolderPrefix = './'
 buildPathPrefix = '/bin/Release/'
 
 packageFolderPrefix = '../'
```
Drigax: This and the project folder prefixes will have to be changed to align with the source code pathing used in L25

3ds Max/Max2Babylon/Max2Babylon_Package.py
Jaskar: My bad ! I didn't test correctly

3ds Max/Max2Babylon/Max2Babylon_Package.py
Jaskar: Yup. I originally wrote a script on another folder, then moved the code when I realized that another script was there.

Jaskar:
This should be good !

Drigax:
Thanks, looks great! Checking in.

Issue #522, opened 2019-05-25
Add Clearcoat support


echadwick-wayfair:
I’d like to see support in Max2Babylon for the new Clearcoat features, via the Physical Material.

Possible implementation design here:
https://forum.babylonjs.com/t/max2babylon-feature-request-clearcoat/3498

Thanks!

Drigax:
I agree. We should make sure our exporters work with our fancy new Babylon.JS features ;)

I'll try to get to this as soon as I'm done putting out fires, definitely sometime this or next month.

sebavan:
Work to be done:

- [ ] Create entities
- [ ] Export max Phys Mat
- [ ] Export max Arnold std Surface
- [ ] Export Maya

sebavan:
Finishing the documentation and I ll be able to close the issue. Please note it is only available in .babylon format as GLTF does not support Coating.

sebavan:
@echadwick-wayfair you can early test from our github prerelease : https://github.com/BabylonJS/Exporters/releases/tag/preRelease20190614.4

all the deployable files are available under assets.

both docs are available:
- Max: https://doc.babylonjs.com/resources/3dsmax#physical-materials
- https://doc.babylonjs.com/resources/maya#physical-materials

PR #542, opened 2019-06-14
Max support for coating

sebavan:
related to #522


Drigax:
Changes look great! Can you upload a .max scene that we can test this against?

PR #544, opened 2019-06-15
Maya and Max PBR Workflow options

sebavan:
Simplify iterations in max by adding a couple of flags to prevent default lighting, allow preprocess env textures and switch to full PBRMaterial for easier troubleshooting.


sebavan:
Also auto Stop/Start Maya on debug build.

Issue #245, opened 2018-06-27
downloader/installer for the exporter


gabrieljbaker:
It would be great to have a more user-friendly installer/executable for this exporter so that the process to get up and running with it is more streamlined. We are directing users to this exporter, but some of them aren't very technical and they would greatly benefit from an installer.

deltakosh:
Excellent point! We will definitely work on that

deltakosh:
We are currently working on this featue

sebavan:
@Jaskar Do we have any ETA to communicate on this feature ? that would be awesome :-)

Jaskar:
Hi !
@sebavan It should be done by the end of the month :)

sebavan:
@Jaskar Thanks a Ton for the ETA !!!

sebavan:
Last missing bits are:

- [ ] Auto Deploy
- [ ] Update install docs

sebavan:
Reference to the doc commit: https://github.com/BabylonJS/Exporters/commit/d85eed74f4c43ae91483db3fcc52a3310ddf6f74

sebavan:
First release available here: https://github.com/BabylonJS/Exporters/releases/tag/preRelease20190617.6

PR #547, opened 2019-06-17
#245 - Windows Installer for 3ds Max and Maya exporters. v0.2

Jaskar:
https://github.com/BabylonJS/Exporters/issues/245


sebavan:
Pretty Good !!!!

Would it be possible to support H-DPI screens ???

Thx,

PR #548, opened 2019-06-19
#245 - Windows Installer for 3ds Max and Maya exporters. v1.0

Jaskar:
- Log / warn / errors are now colored for a better visibility
- An error is raised if the app is not launched in admin mode
- Buttons for "Update all" et "uninstall all"
- "Update all" bug fix
- Update button greyed if already up to date


sebavan:
Looks good :-)

@Jaskar could we use that : https://stackoverflow.com/questions/2818179/how-do-i-force-my-net-application-to-run-as-administrator

to ensure the admin mode is set ???

This would be great !!!

sebavan:
Also, would it be possible to rename form1 to mainForm in the sources ?

sebavan:
@Jaskar, it is almost all good just miss 2 things:
* Rename form1 to mainForm
* make the window a bit larger due to high dpi screen overlaps:
image

I ll merge it right after :-)

sebavan:
Also from the latest tests, it looks like the update is not hidden once installed.

Jaskar:
@sebavan 
- Rename form1 to mainForm -> OK
- make the window a bit larger due to high dpi screen overlaps (hours overlaps button) -> Yep, it's ok. I tried to make it as smaller as possible for little screens. (but I'm not a pro to design forms...)
- Also from the latest tests, it looks like the update is not hidden once installed. -> The update button ? It's working here ... It's hidden only if your exporter is up to date

sebavan:
Ok then I will merge and check on the deployed installer :-)

Jaskar:
Woops, I didn't commit before your merge ! I was waiting for your answer first

Issue #550, opened 2019-06-20
Exporter Installer Neats


sebavan:
I can still experience the UI overlap:
![image](https://user-images.githubusercontent.com/16655165/59846398-0b76f680-9360-11e9-9244-126b71eb46e7.png)

and also the update button is still visible after install:
![image](https://user-images.githubusercontent.com/16655165/59846430-277a9800-9360-11e9-87e2-05efe08ed43b.png)

Both pictures have screen shoted after update.

Jaskar:
1601 is for file not found. I fixed it -> Maya2Babylon.nll.dll instead of Maya2babylon.dll. My bad.
PR is coming !

Jaskar:
https://github.com/BabylonJS/Exporters/pull/551

sebavan:
Thanks

PR #551, opened 2019-06-20
Windows Installer for 3ds Max and Maya exporters.

Jaskar:
Windows Installer : Better fit for small screen.
Fix for Maya update (Maya2Babylon.nll.dll instead of .dll only)


Jaskar:
https://github.com/BabylonJS/Exporters/issues/550

Issue #549, opened 2019-06-19
Export causes mesh scale change in gltf/glb animation


BaldSavant:
In Maya, I'm exporting a character rig which holds 3 unrigged, unconstrained props. The character and props animate over 8300 frames, performing various actions, each of which is contained in a clip (babylon calls them animation groups). The props are scaled from and to 0.001 scale to simulate disappearing during certain actions. This happens multiple times for each prop. On one such time, one of my props is being pulled out from behind the character. During this specific animation, and at no other time, the gltf/glb maintains the scale of the object at 0.001, making the object effectively invisible. 

While debugging this issue, I've opened the file in the babylonjs sandbox viewer and was able to navigate to the clip where this happens, and the prop's position in the hierarchy and set its scale to 1, at which point it worked as expected. I've also confirmed the animation exports as expected as an FBX.

As far as I can tell, there's nothing different about this particular clip in Maya, and the entire section where this happens is contained in a single clip (animation group), so I'm suspecting there's some error during the export. 

Any suggestions?

deltakosh:
Pinging @PatrickRyanMS for guidance

PatrickRyanMS:
@BaldSavant is there any way you can share your source file? It would be easiest to debug if we could look at your exact source to see if there is anything specific to your file that may be making the export fail to write the accurate data. 

From what I am reading you have three meshes that have node animation on them mixed with different meshes with skinned animation and you are only experiencing the issue in the node animations, but I want to be sure I understand exactly how your file is built.

BaldSavant:
That's an accurate summation. I figured you'd ask to see the file, so I've asked my bosses if it's ok to share, but I'll have to wait til I hear back. Hopefully won't take too long, but they're off on a business trip.

PatrickRyanMS:
> edit: and to further clarify, not only am I only experiencing the issue in the node animations, it's only happening on ONE of the node animations. All the rest work as expected.

That was my assumption and tells me there is something specific to that node that we are having some difficulty with on export. Please take whatever precautions you need with the privacy around the model including NDAs and we are happy to do what we need to unblock you.

BaldSavant:
Ok, so we can't release the file currently, but what I CAN do is send you a version of the file with just the props. That's where the problem is anyway, so hopefully that'll do just fine. 

https://drive.google.com/open?id=1DDrzVlvDddi9_G0UNvMSTfx_zRG2Vvh5

[...]

PR #552, opened 2019-06-22
Make baked animations get excluded in export if they aren't relevant within group frame range.

Drigax:
Fix for #549, for GLTF export add additional animation relevance check against entire animation channel if the initial check against the frame range isn't relevant


Drigax:
@sebavan this feels somewhat gross, but let me know if I'm on the right track here. I mainly went with this approach to avoid being dependent on our modelling engine, or the Engine->Babylon export process.

PR #556, opened 2019-06-26
calculate "end frame" of animation group, animation group methods exposed to maxscript

elpie89:
this method allows calculate the last key animated comparing all nodes in the animation group and setting the animation "end frame" to the correct value automatically


sebavan:
@Drigax Could you review this one ?

3ds Max/Max2Babylon/Forms/AnimationGroupControl.cs
```diff
@@ -167,6 +170,40 @@ private void confirmButton_Click(object sender, EventArgs e)
             ConfirmPressed?.Invoke(confirmedInfo);
         }
 
+        private AnimationGroup CalculateEndFrameFromAnimationGroupNodes(AnimationGroup animationGroup)
```
Drigax: This should probably return int, rather than AnimationGroup if we ever want to make this reusable.

Drigax: Also, why put this in AnimationGroupControl, rather than as part of [AnimationGroup](https://github.com/BabylonJS/Exporters/blob/master/3ds%20Max/Max2Babylon/Exporter/AnimationGroup.cs) as a utility or class method?

3ds Max/Max2Babylon/Forms/AnimationGroupControl.cs
```diff
@@ -157,8 +157,11 @@ private void confirmButton_Click(object sender, EventArgs e)
             confirmedInfo.FrameStart = newFrameStart;
             confirmedInfo.FrameEnd = newFrameEnd;
 
-            if(nodesChanged)
+            if (nodesChanged)
+            {
                 confirmedInfo.NodeHandles = newHandles;
+                confirmedInfo = CalculateEndFrameFromAnimationGroupNodes(confirmedInfo);
```
Drigax: Here, we override the UI values with whatever value we calculate. I'm not sure if that's good UX.

Can we change this to prepopulate the text field with our "best guess of the end frame" on animation group creation, rather than render the text field useless?

3ds Max/Max2Babylon/Forms/AnimationGroupControl.cs
```diff
@@ -167,6 +170,40 @@ private void confirmButton_Click(object sender, EventArgs e)
...
+                        lastKey = Math.Max(lastKey, node.TMController.PositionController.GetKeyTime(posKeys - 1));
+                    }
+
+                    if (node.TMController.RotationController!=null)
+                    {
+                        int rotKeys = node.TMController.RotationController.NumKeys;
+                        lastKey = Math.Max(lastKey, node.TMController.RotationController.GetKeyTime(rotKeys - 1));
+                    }
+
+                    if (node.TMController.ScaleController!=null)
```
Drigax: We're looking at posRotScale keys for the objects in our scene, but what about morph target weights? Or if the posRotScale of an object that we want animated is driven via IK or something that may not be keyframed?

3ds Max/Max2Babylon/Forms/AnimationGroupControl.cs
```diff
@@ -167,6 +170,40 @@ private void confirmButton_Click(object sender, EventArgs e)
...
+                    {
+                        int scaleKeys = node.TMController.ScaleController.NumKeys;
+                        lastKey = Math.Max(lastKey, node.TMController.ScaleController.GetKeyTime(scaleKeys - 1));
+                    }
+                    decimal keyTime = Decimal.Ceiling(lastKey / 160);
+                    endFrame = Math.Max(endFrame, Decimal.ToInt32(keyTime));
+                }
+            }
+            animationGroup.FrameEnd = endFrame;
```
Drigax: Lets remove this side effect, I don't think we really want to have a "calculation" function change the state of things outside its scope, it'll be cleaner to return endFrame

3ds Max/Max2Babylon/Forms/AnimationGroupControl.cs
PatrickRyanMS: We should never assume the intent of the user. When I am building an animation timeline, there are definite areas in the timeline I never intend on exporting and if my values in the animation group UI are overwritten with calculated values, I will see that as a bug.

We should help the user by populating with our best guess to save them time, but we should never change what the user set in the UI. If the user enters values that are incorrect (I have keys to frame 50 and export to frame 100) they should see half of the clip with no transforms changing rather than assuming they didn't want those extra frame and eliminating them. There may be a valid reason for them to export their animation in this way that we cannot foresee.

3ds Max/Max2Babylon/Forms/AnimationGroupControl.cs
elpie89: ok I will move this to utility

3ds Max/Max2Babylon/Forms/AnimationGroupControl.cs
elpie89: I will move this calculation on a specific a button, at least the user will be able to add many bones and at the end, the time range can be updated manually. is it ok for you?

3ds Max/Max2Babylon/Forms/AnimationGroupControl.cs
elpie89: I don't really see this point, is the exporter able to export an IK bone without exporting the node that constraint the animation?

3ds Max/Max2Babylon/Forms/AnimationGroupControl.cs
elpie89: I'll move it to tools class and the user will force the calculation manually

elpie89:
So following your suggestion I removed the calculate end frame methods,
it is not used right now, I would like to add it in a specific button, but I wait for your suggestion before to proceed
In order to automate the thousands of mini animation group we have in our production, I exposed more functions to maxscript

3ds Max/Max2Babylon/Forms/AnimationGroupControl.cs
Drigax: That's fine with me, as long as the user can always tweak whatever we calculate to their needs, that's my largest concern.

3ds Max/Max2Babylon/Exporter/AnimationGroup.cs
```diff
@@ -256,6 +256,19 @@ public class AnimationGroupList : List<AnimationGroup>
     {
         const string s_AnimationListPropertyName = "babylonjs_AnimationList";
 
+        public AnimationGroup GetAnimationGroupByName(string name)
```
Drigax: You can usually do all this with a single LINQ statement:

animationGroupList.Where(animationGroup => animationGroup.name = name);

since List implements IEnumerable

3ds Max/Max2Babylon/Forms/AnimationGroupControl.cs
Drigax: Good point, something should be driving this that will be keyframed

PR #559, opened 2019-06-28
Implementation of shared skinning of meshes for GLTF exporter.

Drigax:
Adds checking method "MeshesShareSkin" which does indexed comparison of skinning indices and weights with difference threshold,
modified exporter workflow to only create joint and weight accessors for uniquely skinned meshes.
modified skin exporter to only create GLTFskin objects for uniquely skinned meshes as well.


sebavan:
Looks good to me, feel free to merge when you are all good :-)

PR #563, opened 2019-07-05
[3dsMax][C#][PostBuild] Handle using Max 2016 with Max 2015 exporter version

Selmar:
Still requires the developer to change the postbuild event to pass 2016 to the script instead of 2015, but the rest is handled this way


sebavan:
Thanks

PR #565, opened 2019-07-10
Maya node rotation animation export is incorrect.

Drigax:
Addressing Issue in Maya Rotation keyframe export, where keyframes are incorrect. modified animation keyframe export to convert from eulerAngle rotation in degrees to eulerAngle rotation in radians. Modified BabylonVector3.RotationYawPitchRollToRefBabylon docstring to clarify method usage.


deltakosh:
Lol, that is a sad bug :D

PR #566, opened 2019-07-12
Refactor Babylon->GLTF2 exporter to use shared project

Drigax:
This PR removes the engine-specific dependencies from the 3dsMax Babylon->GLTF2 exporter, and moves it to a shared project, Babylon2GLTF intended to be consumed from the 3dsMax and Maya exporters.

Issue #564, opened 2019-07-10
XREF/Containers support is missing, regeneration of handle is causing many problems


elpie89:
I'm exploring the possibility to integrate containers, to use a reference system
after some day of tests, I came to the conclusion that is not possible right now because of the way the AnimationGroup system is serialized
I discussed this with @Selmar and he agrees with me

We cannot use node handles to serialize animation because they are regenerated in many different ways.
Speaking about containers this happens when a container is closed/opened
Node handles are also regenerated when merging/import scenes 
For that situation, we implemented some months ago the JSON system, but this is a temporary solution
The real problem is that we need to assign to each node a unique GUID and I would like to serialize this as a UserDefiniedProperties
What do you think?
To ensure that the GUID saved in the UserDefiniedProperties  in other we will have to register to some max notification system
For example:
we will have to register to NotificationCodes_NodeCloned to generate a new GUID when a node is cloned

I'm doing all of this because in my team we need to use containers but this will be a benefit for other problem in the future.
For example, doing this we could start to think on the possibility to move the serialization from the root node allowing to merge node between scene and keeping AnimationGroup without transferring the data with JSON

I need to solve this problem asap in my team so I already started, and the  retro-compatibility is guaranteed 

Let me know what do you think, or if you know another system to keep an id consistent when a two-scene are merged or a container  is loaded
I'm open for discussion

elpie89:
Actaully register to the notification Create Object should be enough in my opinion
` m_CreateObjectDelegate = new GlobalDelegates.Delegate5(this.AssignGUID);
                GlobalInterface.Instance.RegisterNotification(this.m_CreateObjectDelegate, null, SystemNotificationCode.NodeCreated);`

deltakosh:
I think this works well! Fancy doing the PR?

elpie89:
yes I will do it tomorrow morning probably

Drigax:
I really like this approach because it gives us an avenue to decouple our
babylon->gltf logic from the max->babylon logic. I was running against our
current implementation of AnimationGroup yesterday when working on #537.

Look forward to seeing the PR.

Thanks,
Nick

PR #567, opened 2019-07-15
guid serailization has been moved to user definied properties

elpie89:
the GUIDs dictionary is reinitialized when a container is added and when a scene is opened
for no Node object, the GUID use the appdata chank GUID 
animation group are serialized referencing the GUID instead of the handle
retro-compatibility is granted to files that were using the old serialization system

Some notes.
I cannot test on Max 2016-2017 right now seems like the two notification I added are using a different Delegate signature in older version. I'm not sure on how to handle this

This is jsut a first step to support containers.
The next one will be to move the serialization from the root node to the container itself


elpie89:
#564

elpie89:
I'm trying to add the compatibility to max 2015-2016-2017
I'm stuck here

```
private void OnNodeAdded(IntPtr objPtr, INotifyInfo infoPtr)
{
                IINode n = (IINode)infoPtr.CallParam; 
}
```

in older version of max is 

` private void OnNodeAdded(IntPtr param0, IntPtr param1)`

I can't find a way to get some info from this signature

Any Idea?

sebavan:
Not for me, let s see if @deltakosh  or @Drigax  might have any ideas ???

Also this will probably need a bit of merging pain after #566 goes in :-( sorry for that.

elpie89:
> I'm trying to add the compatibility to max 2015-2016-2017
> I'm stuck here
> 
> ```
> private void OnNodeAdded(IntPtr objPtr, INotifyInfo infoPtr)
> {
>                 IINode n = (IINode)infoPtr.CallParam; 
> }
> ```
> 
> in older version of max is
> 
> ` private void OnNodeAdded(IntPtr param0, IntPtr param1)`
> 
> I can't find a way to get some info from this signature
> 
> Any Idea?

I found a way finally it's already pushed

Drigax:
Thanks, please merge in master before we can check this in.

elpie89:
should be ok now

PR #570, opened 2019-07-22
fix minor issue creating helper node for containers animation

elpie89:
I forgot to automatically add the node with animation infos inside the container in my last commit


Drigax:
oopsie :)

PR #573, opened 2019-07-25
do not optimize animation preference was not loaded opening the expter panel

elpie89:
just a missing lines from the previous refactoring


Drigax:
Thanks for catching that :)

PR #575, opened 2019-07-25
Fix Maya rotation animation quaternion conversion

Drigax:
This is intended to fix an issue where maya rotation animation keyframes are not accurately exported to Babylon. 

Reimplemented EulerAngle->Quaternion conversion to work with all 123 euler rotation sequences that are supported by Maya. Modified the animation keyframe export process to properly reverse rotation order when changing coordinate system from RHR to LHR.


deltakosh:
You need to update the branch from master

Issue #571, opened 2019-07-23
MaxtoBabylon latest update- export error.


Mustard-Design:
When exporting (Export & Run) from 3dsMax 2019 a 404 error (texture not found) occurs when the browser opens (Chrome and Firefox). Both glb and gltf file types fail.
Previous version (10 days ago) work fine on the same scene (Animation groups deleted and created from scratch).

Update. On further examination, neither gltf or glb files export our character correctly. Biped Animation is not converted and no morph targets work.

Drigax:
Interesting, I haven't noticed this in my own testing, but I'll take a look.

Do the files still load when manually added to the babylon sandbox?

Mustard-Design:
Hi Thanks for looking into this. No the files are still broken when dragging into the sandbox.
If I have simple texture (no bitmaps etc) applied to the model, then it exports and runs the sandbox, but all biped animation is corrupted.
If I add the correct textures to the model (simple png and jpgs) then it will not export at all and I get a 404 error.

Drigax:
Can you share your scene files? I'm unable to reproduce the 404 with a simple scene.

Mustard-Design:
I can, they are under NDA and copyright so I'd have to share them securely with you.

Drigax:
Would email suffice?: whoisdrigax@gmail.com

Mustard-Design:
Hi Yes thankyou. 
I will email you the file. Please don't share this as it's sensitive data, under strict NDA.
Appreciate any help you can offer.

Drigax:
Thank you for sharing, this definitely wrong, both our .Babylon export and glTF export seems to fail with our skinned meshes where either the animation of the skeleton or skinning of the child meshes are incorrect.

However, I am unable to consistently reproduce the 404 issue, the only time when that this occurred for me is when I performed an export and run using a file path that doesn't exist on my system:
![image](https://user-images.githubusercontent.com/3952729/61835964-e54ae780-ae32-11e9-8ee6-1d6519230675.png)

Mustard-Design:
No problem, it seems  skinned mesh +morphs are not working at all right now.
Also I’ve just created a simple export for a sphere and I’m still getting 404 errors.
I’ve attached a screengrab and the 3dsmax file.(2019)


I think the issue is that the exporter is assigning a material name that does not correspond to the actual bitmap reference maybe?
[...]

PR #581, opened 2019-08-02
Remove glTF export texture renaming

Drigax:
As a result of refactoring via #566, I improperly took changes introduced via https://github.com/BabylonJS/Exporters/commit/f97c29c0b12ffbbf736fa19f765eb678f0dba75c into the shared exporter.

As a result, we rename our textures in the Babylon->glTF export, which can negatively affect an asset pipeline due to duplication of textures. 

This appears to be related to #571, where texture paths are broken in exported scene.

This PR removes texture renaming in glTF export, preserving the original texture name provided to the exporter.

### DIFF
--- a/SharedProjects/Babylon2GLTF/GLTFExporter.Material.cs
+++ b/SharedProjects/Babylon2GLTF/GLTFExporter.Material.cs
@@ -335,17 +335,13 @@ namespace Babylon2GLTF
                         //export textures
                         if (baseColorBitmap != null || babylonTexture.bitmap != null)
                         {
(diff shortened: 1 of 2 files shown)


sebavan:
Could you double check that the colon are replaced by underscore in the babylon format just in case :-) as I remember this was the root cause of a handful of issues ?

Drigax:
It looks like your change here: https://github.com/BabylonJS/Exporters/commit/e9ef43403057435062233f1a8f012862d5a5de5c should fix for all cases in the Babylon export.

Issue #537, opened 2019-06-10
Revamp the exporter UIs and add the missing requested export flags support


sebavan:
For both Max/Maya, redesign the exporter UIs.

This should help addressing :

- [x] Initial design exploration for feasibility
- [x] GLTF exporter should be shared across max and maya.
- [x] BackFace Culling #529 
- [x] Double Sided #529
- [x] Transparency Mode #535
- [x] Max Simultaneous Lights #501 
- [x] Custom Attributes #476 
- [x] Animation Only #525 
- [x] Add support for 3dsmax 2020 #618

All the above issues have been grouped here to simplify the development of the new UI.

Do not hesitate to add to the list.

Drigax:
#566 has been checked in, decoupling the Babylon->GLTF exporter from the 3dsMax Babylon exporter implementation.

All that's left to do for 2) is to refactor Maya's exporter to use the shared implementation.

Drigax:
With #593 checked in we can finally check off 2)!

noalak:
I'm currently implementing a solution for the user to specify the extra material attributes in Maya (backface culling, transparency mode, max simultaneous lights and unlit).
Stay tunned for any progress!

jamespedid:
Is custom attributes a low hanging fruit that might be able to be taken out quickly? This would provide tremendous value to me. I'd like to contribute if possible, but I don't have either 3D modeling tool.

deltakosh:
Hey @jamespedid ! @noalak is making a lot of progress and should be able to do a PR really soon.
Form custom attributes, help would definitely appreciated but we cannot help on the 3D modeling license unfortunately

noalak:
We are working on the export animation only feature (#525).
This issue is related to #632.

The 3dsMax exporter is updated to have 2 extra parameters: ExportAnimations and ExportAnimationsOnly. Those parameters drive the export process in order to export only the geometry and then in another export animation(s).

Find attached an example of the resulting babylon and glTF files [elf.zip](https://github.com/BabylonJS/Exporters/files/3878992/elf.zip).

Notice that the glTF animation files use an extension named "AnimationsOnly".
This is the only one reference to an extension, as opposed to what is discussed by @Selmar here https://github.com/KhronosGroup/glTF/issues/1448.
The reuse of an animation (sampler) is supposed to be driven by the channels array according to glTF specs. So we don't see the interest of having an extension specified in each animated node. Maybe I am missing something though.
[...]

PR #583, opened 2019-08-05
Export containers

elpie89:
with the same logic of the flatten hierarchies,
it should be possible to export inherited containers and their animations.
I added a preprocess step that 

- hold the file
- merge the containers.
- load aniamtion group of containers
- export

Of course, everything has been adapted to single and multi export

Please evaluate this PR after the flatten hierarchy one, as his it's a direct consequence


Drigax:
Can you please merge master into this branch as well? Right now its a bit difficult to review since your flatten hierarchy changes are in the PR.

3ds Max/Max2Babylon/Exporter/BabylonExporter.cs
```diff
@@ -179,19 +179,28 @@ public void FlattenHierarchy(IINode node)
...
+                bool merge = containerObject.MergeSource;
+            }
+            AnimationGroupList.LoadDataFromContainers();
+        }
+
+        public void Export(MaxExportParameters exportParameters)
         {
             this.exportParameters = exportParameters;
             IINode exportNode = null;
             if (exportParameters is MaxExportParameters)
```
Drigax: If we're changing the header to use the derived class, then remove this check as well.

I originally opted to use the base ExportParameters class for Export because a goal of #537 was to have both Max and Maya share the same exporter form.

Drigax:
Great, thanks. Did a rudimentary test using animated meshes inherited into a shared scene. Looks like this works well, I just want to make sure we don't cause a null ref depending on who's calling us.

3ds Max/Max2Babylon/Exporter/BabylonExporter.cs
elpie89: ok no problem

PR #580, opened 2019-08-01
flatten hierarchies feature implemented trough hold/fetch logic

elpie89:
so the logic here is 
hold the file
evaluate all hierarchies to check if they are collapsible or not
collapse hierarchies
fetch the file 

#557

PR #584, opened 2019-08-06
Multi export texture folder definition

elpie89:
Added the possibility to define different textures folder for each ExportItem in the multiFileExporter menu


Drigax:
Since this contains  changes on top of #583, it would probably be best to review this after that is checked in.

elpie89:
i'l merge the master in here tomorrow morning

Drigax:
Thanks, your changes look good!

elpie89:
As I was working on path features I took the time to polish the format path utilities
Our artists were asking for a better FolderBrowser, the default one of WinForms is really bad, I added the Microsoft.WindowsAPICodePack to the project to use an enounced folder browser

Also, the old way to distinguish the relative paths and the absolute paths weren't clear
I replaced with a system that use colors to give feedback

so this :
`[project folder]relativePathToGltf` 
is now just an absolute path formatted with two different colors

This way the entire format unformat logic is easier and I removed some unused methods
I think this is enough for this PR

Drigax:
@elpie89 I believe [your latest change to this PR](https://github.com/BabylonJS/Exporters/pull/584/commits/dbe0444edae1b64722f2c375a350f0202d7047c7) caused a regression (#587).

To quickly unblock our users, I'll need to revert this specific commit. Please create a new PR with the commit so we can figure out why this is breaking our installation.

Drigax:
Sorry, I was wrong to revert this change. It turns out this was caused by the reference added to ManagedServices.dll via #580. I was able to fix our installer, and will revert my reversion shortly.

Issue #585, opened 2019-08-07
Multi Export Layer based


elpie89:
Artists are asking to be able to export many files simultaneously based on multiple layer selection
This feature is very useful when you have a scene with multiple LODs for an asset and you need to export everything at the same time 

I can work on this feature, you can assign it to me

Drigax:
Sounds good! thanks.

PR #601, opened 2019-08-23
Export layer based

elpie89:
#585 
Added the possibility to select which layer to export on the multi export panel.
This is particularly useful in case of export of many LODs 

The export items were saved in the root node using the node handle.
I replaced this with a system based on GUID to be able to define more export item for the root node


elpie89:
The installer needs to be updated because I added to the project some new dll of the max SDK
https://help.autodesk.com/view/3DSMAX/2017/ENU/?guid=__files_GUID_A08A117C_9D15_4044_BC29_3AD9F8AFD9FE_htm

deltakosh:
Please be patient. @Drigax will be back on Monday :)

elpie89:
> Please be patient. @Drigax will be back on Monday :)

no problem, no rush

3ds Max/Max2Babylon/Forms/MultiExportForm.cs
```diff
@@ -216,6 +250,61 @@ private void ExportItemGridView_CellDoubleClick(object sender, DataGridViewCellE
...
+
+
+            // double-clicked layers column cell, select some layers!
+            if(e.ColumnIndex == ColumnLayers.Index)
+            {
+                if (sceneExplorer==null )
+                {
+                    sceneExplorer = new SceneExplorerDialog("Layer to Export");
+                    sceneExplorer.ExplorerControl.EditingEnabled = false;
+                    sceneExplorer.Closed += SceneExplorerOnClosed;
```
Drigax: This doesn't seem to work very well. There's no UI to confirm your layer selection, and closing the scene explorer doesn't appear to select the layers as well. Also, selecting the objects instead of the layers that you want to export doesn't make for a very good experience as well.

image


@PatrickRyanMS currently this proposes that double clicking a Layer cell opens the scene explorer to select objects, and the layers of those selected objects should be added to the multi-export item.

elpie89:
I'll do a second pass to improve the experience, adding a confirmation dialog, but it is really complex to manage 3dsmax native UI
In this case, I wanted to display only layers but I can't find any way.
Let's leave it like this for now without accepting pause the pull request, I will improve it soon

elpie89:
This one should be ok, the layer Selector is based on a WinForm TreeView and could be easily managed in the future.
There is still some glitch but generally speaking, the UX is improved.

Drigax:
Great, for reference I was able to test this using this simple scene here:
[MaxLayerExport.zip](https://github.com/BabylonJS/Exporters/files/3571131/MaxLayerExport.zip)

I especially appreciate the simple winform checkbox list, this definitely seems like an improvement over the native scene viewer window's capabilities.

If you can merge master, I'll get this in.

PR #603, opened 2019-08-26
support for multiple container linked to same source

elpie89:
This feature allows importing multiple times  the same containers, 
on import in ID as a suffix is added
on export, the exporter resolve GUID of aniamtionGroup.nodes id when the container is opened


3ds Max/Max2Babylon/Exporter/AnimationGroup.cs
```diff
@@ -522,6 +525,66 @@ public static void LoadDataFromContainers()
...
+            }
+
+            //replace animationList guid to have distinct list of AnimationGroup for each container
+            string animationListStr = string.Empty;
+            container.BabylonContainerHelper().GetUserPropString(s_AnimationListPropertyName, ref animationListStr);
+            string[] animationGroupGuid = animationListStr.Split(AnimationGroup.s_PropertySeparator);
+            foreach (string guidStr in animationGroupGuid)
+            {
+                Guid newAnimGroupGuid = Guid.NewGuid();
+                helperPropBuffer = helperPropBuffer.Replace(guidStr, newAnimGroupGuid.ToString());
```
Drigax: Looks like my test scene is causing a null ref here where we retrieve an empty animationList string.

Drigax:
Ah, good find. I was able to reproduce it using this series of scenes:

Where our first imported container is animated in both animation groups, even though Shield_Container001 is selected for the second group.

[ContainerImportTest-MultipleImportOfContainer.zip](https://github.com/BabylonJS/Exporters/files/3557558/ContainerImportTest-MultipleImportOfContainer.zip)

Testing against this scene we nullref at L556 of AnimationGroup.cs

elpie89:
the null reference exception has been managed.
This error happened trying to export a container without previously saving Animations into Babylon animation helper
In that case, a default one is created without animation and a message has been added to notify the user that he should do this on the container source before the export stage

Drigax:
Great, please merge master, and I'll merge.

PR #607, opened 2019-08-31
Fix null reference when exporting material with no alpha map and no base color map.

Drigax:
Fix to nullreference of baseColorTexture file extension, reorganized export code for clarity.


sebavan:
LGTM, feel free to merge if all tests scene pass.

Issue #606, opened 2019-08-29
Add support for TDD


elpie89:
Adding features is becoming more and more complex, we should add some sample scene and implement a proper Test-Driven development pipeline

Drigax:
I completely agree. I think we can quickly get some test coverage via an integration testing approach (Similar to our CI pipeline for Babylon.JS):

- [ ] Itemize functionality and create individual test cases for functionalty

- [ ] Create harness for batch export of test scenes

- [ ] Create library of per-testcase validation images 

- [ ] Create harness for loading export in trusted viewer and performing per-pixel comparison with validation image

- [ ] Create simple harness for doing 2 & 4

Eventually:

- [ ] Setup Azure instance to run #5 for pull requests

deltakosh:
Up?

deltakosh:
Removing 4.2 and flagging it as future

Drigax:
I definitely think this should be a long-term goal of this project. via #969 we're starting to run into more concerns about regressing behavior, and without some regression testing solution, we'll never know if our changes are breaking some known behavior.

PR #616, opened 2019-09-10
Fix skin

elpie89:
This fix is really important we still had the following problem.
A mesh skinned to a non root bone, doesn't export well
This is the test file I'm using, I suppose we can keep this file for #606 
[skin_test_sample.zip](https://github.com/BabylonJS/Exporters/files/3596222/skin_test_sample.zip)
With this pull request the logic will work for gltf extension
but I broke something in the babylon extension
Basically from what I'm seeing and what I'm understanding we have a conflict
gltf need a node hierarchy and a list of minimal list of bone in the skin list 
the gltf get the skin position using his reduced bone list and the legacy node list

Here I refactored splitting the exportNodeRec  in exportNodeRec and exportNodeTypeRec
I think we need to export GLTF nodes before, then evaluate skin, camera etc.

The babylon format need the entire bones hierarchy including nodes that artist didn't specify as part of the skin
Am I right?
I do not expect that this pull request is accepted as is, but if someone work on top of this to fix the babylon format this pr could be accepted


elpie89:
of course this broke MayaExporter too.
@Selmar @Drigax I need some help on this, can you take a look ?
Thanks

SharedProjects/Babylon2GLTF/GLTFExporter.Skin.cs
```diff
@@ -4,6 +4,8 @@
 using System;
 using System.Linq;
 using System.Collections.Generic;
+using Autodesk.Max;
```
Drigax: Why are we referencing Max components in the shared code?

SharedProjects/Babylon2GLTF/GLTFExporter.Skin.cs
```diff
@@ -153,29 +156,31 @@ private GLTFSkin ExportSkin(BabylonSkeleton babylonSkeleton, GLTF gltf, GLTFNode
...
             }
 
-            var boneNodePair = boneToGltfNodeMap.FirstOrDefault(pair => pair.Key.id.Equals(babylonBone.id));
-            if (boneNodePair.Key != null)
-            {
-                return boneNodePair.Value;
-            }
 
-            // Node
-            var gltfNode = new GLTFNode
```
Selmar: if I'm not mistaken we never get here anymore, every node should be in the nodeToGltfNodeMap already

i think this whole function may be unnecessary

@drigax is there a case where a node or bone may be unrepresented by a BabylonNode, or does every node/bone have a matching BabylonNode?

SharedProjects/Babylon2GLTF/GLTFExporter.cs
```diff
@@ -93,7 +93,15 @@ public void ExportGltf(ExportParameters exportParameters, BabylonScene babylonSc
             {
                 exportNodeRec(babylonNode, gltf, babylonScene);
                 progression += progressionStep;
-                logger.ReportProgressChanged((int)progression);
+                //logger.ReportProgressChanged((int)progression);
```
Selmar: of course this progression stuff will have to be updated

Selmar:
I do not see any changes that modify the babylon export, I assume the exporter was already broken for this file (@elpie89 could you verify that?). Perhaps I'm overlooking something.

elpie89:
> I do not see any changes that modify the babylon export, I assume the exporter was already broken for this file (@elpie89 could you verify that?). Perhaps I'm overlooking something.

You are right, the result was already broken exporting the test scene
I wrongly assumed that was my fault

SharedProjects/Babylon2GLTF/GLTFExporter.Skin.cs
elpie89: it was an unused namespace reference, sorry I forget it there.
It has been removed now

Drigax: Cool, we fixed the build break. :)

SharedProjects/Babylon2GLTF/GLTFExporter.Skin.cs
elpie89: so I confirm that this method is not necessary
I comment out the method and his call.

elpie89:
I found the problem 
when we calculate the matrix of the skinned bones, we have to get the world matrix of the root bones
` matrix = (parentIndex==-1)?node.GetWorldTM(0).ToArray():node.GetLocalTM(0).ToArray()`
I think this can be merged now

SharedProjects/Babylon2GLTF/GLTFExporter.Skin.cs
Drigax: I don't think that this is true at all.

See this scene:
[SimpleSkinningCylinder.zip](https://github.com/BabylonJS/Exporters/files/3617813/SimpleSkinningCylinder.zip)

When exported, we have a babylon scene containing one node, the mesh that is skinned, while the bones are exported as part of the babylon skeleton. It appears that we consider skinning bones separately from nodes in this format, while in glTF skeletons are composed of nodes

bghgary: > glTF skeletons are composed of nodes

Yes, I can confirm that glTF skin joints point to node in the scene hierarchy.

elpie89:
so practically the idea is just to remove the BoneToGLTFNode map
initializing the scene every BabylonBone is converted to a BabylonNode on the GLTFExporter side
tested with all test with our initial scene and with @Drigax scene

elpie89:
another test scene 
[test_scale_GLTF.zip](https://github.com/BabylonJS/Exporters/files/3738055/test_scale_GLTF.zip)

elpie89:
@Drigax I do not understand if there is any problem here.. are you in? Because after last modification everything works on our side, and it has been tested with many basic files and in production as well

Selmar:
Since we're changing the output of Max2Babylon, I think a number of these changes also need to happen in the Maya exporter, correct? For example, the removal of `-bone` if gltf is exported.

We do not work with Maya at all, however, so we can't do this for you.

The conversion from bones to nodes is on the shared project side, fortunately.

Drigax:
@elpie89, sorry, I've been a bit busy with other projects to take a good look at this. I've been mostly waiting for your code to stabilize before I test correct behavior in Max ( Most likely later today or tomorrow, since i've been putting this off long enough). 

And I'll handle any Maya related regression after checking in. I can't tell off hand, your change looks pretty universal.

Thanks for all your help,
Nick

Issue #620, opened 2019-09-12
Export Layer features improvement


elpie89:
- the export layer based should not be an exclusive of the multi-exporter (lowpriority)
- if a user specify one or more layer to export this should be forced to be visible
many times artist select a layer that is hidden expecting to see the content in the file

PR #622, opened 2019-09-13
Layer exporter improvements

elpie89:
as explained in #620 
I haven't linked the layer export feature to the single export page.
It is useless and I will not do it probably.

I prefer to work on something more important.

For this PR I just made the layer visibility management features 
and made a polish pass on the whole Layer exporter feature.


Drigax:
Looks good on my end, used the same scene as #601, but with hidden layers.

PR #623, opened 2019-09-13
add 3ds max 2020 support !!Update 2 mandatory!!

susares:
It is important that this only works with Update 2 installed. Otherwise 3DS Max will crash after hitting the export button. This is due to some critical internal .NET SDK bugs that have been existing since original 3ds Max 2020 release and were fixed in the latest update.


Drigax:
Great work! I'll review this first thing Monday

Drigax:
I'll have to wait until my network deployment makes Update 2 available before I can properly test and review this.

PR #624, opened 2019-09-16
Optimization

elpie89:
I made a pass of optimization.
Animation group were looking for the node by GUID traversing all the scene over and over
Now this happens only if the node is not found in the GUID dictionary

This is the result before and after on two big files.
Consider that there are almost 500 animation groups for each file
There is still a lot of work to do especially exporting nodes, but this is just a first step
image

### DIFF
--- a/SharedProjects/Babylon2GLTF/GLTFExporter.cs
+++ b/SharedProjects/Babylon2GLTF/GLTFExporter.cs
@@ -33,6 +33,8 @@ namespace Babylon2GLTF
             this.logger = logger;
 
             logger.RaiseMessage("GLTFExporter | Exportation started", Color.Blue);
+            var watch = new Stopwatch();
+            watch.Start();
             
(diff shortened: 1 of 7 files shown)


elpie89:
with last fix the time spent on animation group export I think is acceptable.
this is the last report made on the second file (the biggest one)

image

SharedProjects/Babylon2GLTF/GLTFExporter.cs
```diff
@@ -83,6 +87,10 @@ public void ExportGltf(ExportParameters exportParameters, BabylonScene babylonSc
                 logger.ReportProgressChanged((int)progression);
                 logger.CheckCancelled();
             }
+#if DEBUG
+            var t1 = watch.ElapsedMilliseconds / 1000.0;
```
Drigax: Can we please use more descriptive variable names than t1-t4?

PR #630, opened 2019-09-18
Babylon Exporter Installer v1.2.0 - Add support for Max 2020

Drigax:
Added 2020 installer workflow to main menu, modified max panel to display scroll bar


Drigax:
@Jaskar your design for the updated made adding support for future versions a breeze :) thanks!

Jaskar:
Could be more automatic imho, but I tried to make it simple ! :)

PR #631, opened 2019-09-19
removed guid suffix for gltf bones

elpie89:
this suffix was only for babylon exporter until last month
It was causing node duplication on the exporter gltf
[skin.zip](https://github.com/BabylonJS/Exporters/files/3630129/skin.zip)
I used this file to test it

### DIFF
--- a/3ds Max/Max2Babylon/Exporter/BabylonExporter.Skeleton.cs
+++ b/3ds Max/Max2Babylon/Exporter/BabylonExporter.Skeleton.cs
@@ -309,7 +309,7 @@ namespace Max2Babylon
                 // create the bone
                 BabylonBone bone = new BabylonBone()
                 {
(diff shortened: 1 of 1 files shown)


Drigax:
I originally did this due to https://github.com/BabylonJS/Exporters/pull/578, which i admit was probably the wrong way to go about fixing this.

Similarly to #616, changing our bone naming convention affects how we query bones for animation group inclusion:
https://github.com/BabylonJS/Exporters/blob/61eee8aa9e51dd54ab38b60dbfbfed41cbccfdba/3ds%20Max/Max2Babylon/Exporter/BabylonExporter.Animation.cs#L42
and 
https://github.com/BabylonJS/Exporters/blob/61eee8aa9e51dd54ab38b60dbfbfed41cbccfdba/3ds%20Max/Max2Babylon/Exporter/BabylonExporter.Animation.cs#L76-L84

In my limited investigation, making the change to L72 to only add the -bone guid suffix when exporting .babylon appears to fix our skeletal animation export, but breaks your test scene.

elpie89:
this can be closed as is solved with #616

Issue #628, opened 2019-09-16
3dsMax - Add Animation Baking


Drigax:
Deliverable for #619.

- [x] Similar to the [Maya implementation](https://github.com/BabylonJS/Exporters/blob/master/Maya/Exporter/BabylonExporter.Animation.cs#L56-L165), it would be a great addition to the 3dsMax exporter to add the option to bake animations per-frame.

- [ ] an additional deliverable would be to use the baking functionality to bake the first and last frames of an animation group at export, as requested via #596

elpie89:
I need to spend time on #632 then I can work on this

Drigax:
removing the second deliverable from this task, it's already best tracked via #596.

PR #635, opened 2019-09-26
Animation baking

elpie89:
as discussed in #628 
In my team, artists asked also to be able to select which node should be baked or not
So I added an Enum selection in the export
Trough an action, in the quad menu as  for the flatten feature, the user is able to mark a node as To BakeOrNot


Drigax:
Great work, I'll take a look at this first thing tomorrow. Got an idea of a good way to test this? I assume making an object animate in some circular path with and without baking should show the difference since there's no easing implementation in glTF yet

Drigax:
Hey, in testing this, I noticed that we seem to already bake some of our keyframe animation export in 3dsMax:

https://github.com/BabylonJS/Exporters/blob/3f10b21872af96943f50c69f10dd162e5ed8df6b/3ds%20Max/Max2Babylon/Exporter/BabylonExporter.Animation.cs#L555-L567

I created this simple scene to test you contributions and noticed that even with baking disabled the easing of the animation tracks is preserved.
(originally I assumed that we exported the raw keyframe data without tangents, and the consuming engine interpolated linearly)

I'm wonder what's the best way to reconcile your additions, maybe we mis-characterized the problem here.

@elpie89, What functionality is currently lacking for your artists that your scene pre-processing approach fixes? Also, do you have a test scene that demonstrates this? It would be great to know what the current shortcoming with the existing approach are.

I think the per-object attribute for animation baking is a great addition as well, and is a good way to expose this to the user. I'm sure if we don't actually need the preprocessing step, we could implement keyframe retrieval, and use this property to toggle between those animation export functions (Similarly to how we handle [GetAnimation](https://github.com/BabylonJS/Exporters/blob/3f10b21872af96943f50c69f10dd162e5ed8df6b/Maya/Exporter/BabylonExporter.Animation.cs#L172-338) and [GetAnimationsFrameByFrame](https://github.com/BabylonJS/Exporters/blob/3f10b21872af96943f50c69f10dd162e5ed8df6b/Maya/Exporter/BabylonExporter.Animation.cs#L418-456) in Maya).

3ds Max/Max2Babylon/2017/BabylonToogleBakeAnimation.cs
```diff
@@ -0,0 +1,80 @@
+﻿using System.Windows.Forms;
```
Drigax: Can we rename this to ToggleBakeAnimation?

3ds Max/Max2Babylon/Exporter/BabylonExporter.cs
```diff
@@ -179,6 +179,62 @@ public void FlattenHierarchy(IINode node)
...
+
+            if (bakeAnimationType == BakeAnimationType.BakeAllAnimations)
+            {
+                foreach (IINode n in Tools.ITabToIEnumerable(tobake))
+                {
+                    n.SetUserPropBool("babylonjs_BakeAnimation", true);
+                }
+            }
+
+            ScriptsUtilities.ExecuteMaxScriptCommand(@"
```
Drigax: As I was saying in my previous comment, this pre-pass may not be necessary, since we retrieve each node's transform each frame for our animation.

3ds Max/Max2Babylon/Exporter/MaxExportParameters.cs
```diff
@@ -3,13 +3,23 @@
...
+        BakeSelective
+    }
+
+
     public class MaxExportParameters : ExportParameters
     {
         public Autodesk.Max.IINode exportNode;
         public List<Autodesk.Max.IILayer> exportLayers;
         public bool usePreExportProcess = false;
+        public bool keepChangement = false;
```
Drigax: is keepChangement "Apply preprocessing effects to scene permanently"?

elpie89:
I will handle this tomorrow

3ds Max/Max2Babylon/Exporter/MaxExportParameters.cs
elpie89: yes I think is useful when many preprocess are applied and a user wants actually see what BabylonExporter is doing to his scene.
But this is not permanent as the scene is not saved and just reloading the scene manually a user can get back hi scene
Should I rename it?

elpie89:
> Hey, in testing this, I noticed that we seem to already bake some of our keyframe animation export in 3dsMax:
> 
> https://github.com/BabylonJS/Exporters/blob/3f10b21872af96943f50c69f10dd162e5ed8df6b/3ds%20Max/Max2Babylon/Exporter/BabylonExporter.Animation.cs#L555-L567
> 
> I created this simple scene to test you contributions and noticed that even with baking disabled the easing of the animation tracks is preserved:
> 
> [TestAnimationBaking_followPathWithEasing_BakeVsNotBake.zip](https://github.com/BabylonJS/Exporters/files/3664909/TestAnimationBaking_followPathWithEasing_BakeVsNotBake.zip)
> 
> (originally I assumed that we exported the raw keyframe data without tangents, and the consuming engine interpolated linearly)
> 
> Which makes me wonder what is the best way to reconcile your additions, maybe we mis-characterized the problem here if we already bake TRS and morph weight animations.
> 
> @elpie89, What functionality is currently lacking for your artists that your scene pre-processing approach fixes? Also, do you have a test scene that demonstrates this? It would be great to know what the current shortcomings with the existing approach are.
> 
> I think the per-object attribute for animation baking is a great addition as well, and is a good way to expose this to the user. I'm sure if we don't actually need the preprocessing step, we could implement raw keyframe retrieval, and use this property to toggle between those animation export functions (Similarly to how we handle [GetAnimation](https://github.com/BabylonJS/Exporters/blob/3f10b21872af96943f50c69f10dd162e5ed8df6b/Maya/Exporter/BabylonExporter.Animation.cs#L172-338) and [GetAnimationsFrameByFrame](https://github.com/BabylonJS/Exporters/blob/3f10b21872af96943f50c69f10dd162e5ed8df6b/Maya/Exporter/BabylonExporter.Animation.cs#L418-456) in Maya).

So all the goal of this PR is to also be able to export constrained nodes, not to only convert animation curves.
Right now the Babylon Exporter does not resolve constraint between two object
Here a sample scene
[BakeConstraintBased.zip](https://github.com/BabylonJS/Exporters/files/3681017/BakeConstraintBased.zip)

Looking on the node named "target" it is possible to see that without animation bake the animation is completely wrong/missing

3ds Max/Max2Babylon/Exporter/MaxExportParameters.cs
Drigax: Makes sense, can we name this to "ApplyPreprocessToScene" or something similar? I just want to make sure that this option is clear to the user.

PR #642, opened 2019-10-04
AnimationGroup support xrefObject and merged scene

elpie89:
As with containers on source scene in the action menu StoreAnimation groups
Merge/link scene in a master scene
Click on action menu Load Animation Groups
If an animation group is already there with the same name, the two animation groups are merged


Drigax:
Thanks, can you please resolve the merge?

PR #654, opened 2019-10-14
Exporter extension

elpie89:
The idea is to be able to add a sort of hook in the GLTFexporter giving the possibility to add a custom extension
Right nows there is no possibility to do this except forking and modifying the source code or eventually replacing the entire per Type Exporter logic as we did for the material exporter
A typical use case could be to add an extension on Gltf to define a mesh as a collision
In this case, we should be able to define the extension in another class and add the extension without rewriting or modifying, for instance, the GLTFExporter.Mesh file.

The result will be something like this
```"meshes": [
        {
            "primitives": [
                {
                    "attributes": {
                        "POSITION": 1,
                        "NORMAL": 2,
                        "TEXCOORD_0": 3
                    },
                    "indices": 0
                }
            ],
            "name": "Teapot001",
            "extensions": {
                "my test": {
                    "test": 2.0,
                    "index": 0
                }
            }
        }
    ]
```
I add a sample class to verify and play with it
[CustomExtension.zip](https://github.com/BabylonJS/Exporters/files/3724718/CustomExtension.zip)


Selmar:
There are some distinctions to be made between different types of exporter extensions. There are two types we currently have:
- Define how to export an item (like the material exporter extension)
    - Tied to 3dsmax ClassID.
    - This can be used to define how an object should be exported (say, an editable poly or skin). With some modification, this pattern could be used to export lights, which are a completely new object (it's `KHR_lights_punctual`, if I'm not mistaken?).
- Add something to the exported item (mesh defining collision objects)
    - Tied to babylon type.
    - This is to add (gltf) extension data to an object type already exported by babylon, such as collision data for meshes.

For animation retargeting I think this implementation is fine. For defining new animation channels, I am not sure if the above approach suffices. We'll see.

Drigax:
I see. I'll merge this in as is. @elpie89, can you please resolve the build failure? There was a merge conflict that I resolved, and forgot to remove the animation group modifications you made to bone export via #616.

PR #656, opened 2019-10-17
Bugfix generic: Xref, Multi Export

elpie89:
Exporter wasn't able to merge multiple Xref
Exporting multi export (layer based) we should force children of a layer to be visible

### DIFF
--- a/3ds Max/Max2Babylon/Exporter/BabylonExporter.cs
+++ b/3ds Max/Max2Babylon/Exporter/BabylonExporter.cs
@@ -243,6 +243,7 @@ namespace Max2Babylon
             foreach (IIContainerObject containerObject in sceneContainers)
             {
                 if (!containerObject.IsInherited)continue;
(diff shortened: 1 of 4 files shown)


3ds Max/Max2Babylon/Exporter/BabylonExporter.cs
```diff
@@ -243,11 +243,26 @@ public void ExportClosedContainers()
             foreach (IIContainerObject containerObject in sceneContainers)
             {
                 if (!containerObject.IsInherited)continue;
+                ScriptsUtilities.ExecuteMaxScriptCommand($@"(getNodeByName(""{containerObject.ContainerNode.Name}"")).LoadContainer()");
+                bool update = containerObject.UpdateContainer();
```
Drigax: 3ds Max\Max2Babylon\Exporter\BabylonExporter.cs(247,47): Error CS1955: Non-invocable member 'IIContainerObject.UpdateContainer' cannot be used like a method.

elpie89: there is probably a compatibility problem because I tested only with max 2019
I'll fix it tomorrow

Issue #535, opened 2019-06-05
Maya Exporter to support alphaMode MASK and alphaCutoff


vanlicht:
Hi,

Really enjoy the Maya Exporter  but find the feature still missing to support alphaMode MASK and alphaCutoff values. Any solution already exists that I miss? Thanks!

Drigax:
I'm not entirely sure off hand. @vanlicht, can you provide a sample .maya scene to look at to test correctness of the implementation? I think we have a few other alpha channel related issues going off of #473, sounds like they're related.

vanlicht:
Hi @Drigax,

Please see the attached. Thanks.

[gittree.zip](https://github.com/BabylonJS/Exporters/files/3259203/gittree.zip)

Drigax:
Hi @vanlicht, I'm not sure if the difference in your results are due to AlphaMode export settings. Some leaves aren't rendered in your left example due to backface culling of the leaf meshes. Changing the material properties in the Babylon engine, I can see:

![image](https://user-images.githubusercontent.com/3952729/59125647-cdb5af00-8917-11e9-9d9c-24765a0e13d8.png)
Where the left tree's leaves uses a material with backface culling enabled, while the right tree's leaves' material has backface culling diabled.

Which appears to replicate your initial test.

I'll try to create a test rig for testing the AlphaModeMask & AlphaCutoff capabilities, more similar to the the one used in the GLTF spec:
https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/AlphaBlendModeTest

vanlicht:
Hi,
Yes I am aware if the doubleSided issue, which at the moment there is no support from Maya exporter I think. For the desired result I had to manually add it in the glTF code. 
Though, there is difference of the tree in BLEND mode versus MASK mode, wonder if you noticed the difference. Thanks.

Drigax:
Thanks for the clarification. I don't have an eye for that fine of detail, so I wanted to address the most obvious issue first. 😊 

But i can see the need to support removing the leaves' white halo via a hard alpha test. 

Currently, Maya doesn't expose texture Alpha blending modes, nor per-material back face culling in the Maya exporter, and our exporter implementation doesn't currently extend that functionality (most likely due to implementation complexity).

Some goals of our exporter UI revamp would be to expose these options, and ideally using as much Maya & 3dsMax native UI as possible, and I think that it would be best to support this as part of a drastic change rather than hacking in a stopgap to support this right now.

Issue #529, opened 2019-06-04
add backface culling and double sided support in Maya


vortice3D:
Hello,
We usually deal with transparency in our PBR materials, but don't have any way to specify the backfacing culling and double sided attributes later on.
How are these attributes handle, if they are?
Just in case, we would like to have a way to specify these attributes on a per material basis, like in Blender.
Thanks in advance.

Drigax:
Hi @vortice3D. I don't believe that the maya exporter currently supports backface culling. We just added the funcitonality to the 3dsMax exporter, so we should add this to the Maya exporter as well. Unfortunately, you may have to add the material property manually as a workaround for now, but I'll try to get this parity in our Maya exporter in the next few weeks.

Can you give me a simple scene to test the correctness of the implementation?

vortice3D:
Hi there,
I don't know which information do you have for the implementation.
But basically, there should be a way to tell the glTF material to disabled the backface culling attribute when there is any transparency or opacity going on.
In the attached sample file, the cube is red and semi transparent, and it appears as this inside sandbox:
![imagen](https://user-images.githubusercontent.com/1760601/59086704-ccc65280-8902-11e9-81ab-0245de69d3d8.png)

But it should have the backface culling turned OFF and be like this:
![imagen](https://user-images.githubusercontent.com/1760601/59086802-19aa2900-8903-11e9-8197-ed3afbf59a3a.png)

If this is implemented by adding custom attrbitues to the material, that's fine for now (you'll have to explain if that custom attribute should be added to the material itself or to the Shading Group).
I forgot to mention that we use Arnold's aiStandardSurface materials, but with StingrayPBS ones the problem is the same. The backface culling attribute can't be set, so it won't be exported at all.

Aside from that, there should be a way to control the transparency mode also.

[MAYA_SingrayPBS.zip](https://github.com/BabylonJS/Exporters/files/3264772/MAYA_SingrayPBS.zip)

sebavan:
Closing for house keeping as we are grouping all the UI work in one place #537

Issue #501, opened 2019-05-02
Add exporter support for modifying BabylonMaterial.maxSimultaneousLights


Drigax:
We should be able to modify this property at export time to allow users to better preview their .BABYLON scenes without additional scripting.

Drigax:
I should be able to implement this in around 2 months, I want to tackle some higher priority issues first.

sebavan:
Closing for house keeping as we are grouping all the UI work in one place #537

Issue #476, opened 2019-04-13
Export custom Attributes from Maya


Mudoglu:
Is there any way to export custom Attributes from Maya with the Babylon Exporter?

In maya you can asssign Attributes to objects.
We would like to assign a lightmap Attribute to connect later the lightmap texture to the right shader.

https://forum.babylonjs.com/t/export-custom-attributes-from-maya/2518/2

If needed i could provide a scene showing what we like todo.

Cheers

deltakosh:
Thanks a lot, we're gonna add that to the todo :)

deltakosh:
Pinging @PatrickRyanMS FYI I think we should consider it for our UI rework)

kadirilkimen:
I think 3ds max exporter plugin may have similiar function.
In 3ds max, we can append custom attributes to any object ( floats, strings etc ).
It would be very good if babylon exporter append this custom attributes to object's data.

I dont know if it is same task ( because i dont know anything about Maya) , but i think it worth to mention here :)

sebavan:
Closing for house keeping as we are grouping all the UI work in one place #537

PR #668, opened 2019-10-31
Add the missing requested export flags support #537

noalak:
Main issue #537
- BackFace Culling #529
- Double Sided #529
- Transparency Mode #535
- Max Simultaneous Lights #501
- Custom Attributes #476

Double Sided for Maya is still work in progress and is not pushed to this PR.

Some details on how we did it:
- Backface culling, transparency mode and max simultaneous are new attributes attached to materials. In Maya they are attached to a separate node. In 3dsMax they are on native materials directly.
- Custom attributes added by the user to a node or a material are exported. All primitive types are supported (string, boolean, float, int, vector3, enum).
- Double sided native material of 3ds Max is treated as a multimaterial. All mesh faces are duplicated and flipped.

Some feedbacks would be welcomed before merging this PR.


Drigax:
Thanks for the contribution! This PR is a biggun, I'll have to take some time to review it throughly. At a glance the code quality seems quite up to standards, I'm not too concerned with that, but I do want to make sure that I can skim over every change. I'll also make time to functionally test this in accordance with your validation tests. Can you provide the test that you used in the summary you provided, that would definitely help.

Thanks!

noalak:
Sample files can be downloaded here: [https://we.tl/t-WupXcm3veg](url)
They are also contained in this PR, under Samples folder for each project (3dsMax and Maya).

Once you reviewed and tested, we will zip dlls (for direct download) and update the installer for Maya (new files to copy).

Drigax:
@noalak I don't believe that the dll zipping is necessary, we auto-build master after each checkin and publish via this CI pipeline:

https://dev.azure.com/babylonjs/ContinousIntegration/_build/results?buildId=2325

3ds Max/Max2Babylon/Exporter/BabylonExporter.Material.cs
```diff
@@ -1,509 +1,605 @@
...
+
+                if (babylonMaterial.transparencyMode == (int)BabylonPBRMetallicRoughnessMaterial.TransparencyMode.ALPHATEST)
+                {
+                    // Set the alphaCutOff value explicitely to avoid different interpretations on different engines
+                    // Use the glTF default value rather than the babylon one
+                    babylonMaterial.alphaCutOff = 0.5f;
+                }
+
+                // Add babylon attributes
+                AddPhysicalBabylonAttributes(materialNode.MaterialName, babylonMaterial);
```
Drigax: Similarly to the Maya exporter, after testing the workflow, requiring a "dryrun" to add custom attributes feels like a suboptimal experience. Can we look into possible ways to hook into the slate editor and add attributes on node creation? Or, at possibly expose a UI or context menu control that does this?

Maya/AETemplates/AEbabylonAiStandardSurfaceMaterialNodeTemplate.mel
```diff
@@ -0,0 +1,27 @@
+global proc AEbabylonAiStandardSurfaceMaterialNodeTemplate ( string $nodeName )
```
Drigax: (This comment applies to 3ds Max/Samples/sourceModels/BabylonMaterialAttributes/BackfaceCulling.max)

image

I've noticed that this scene raises a corruption warning? Can you look into what may be causing this?

Drigax: It appears related to these scripts possibly being referenced by the scene? They do not exist on my local computer, however...

image

Drigax:
Also, can the source sample scenes be revised to rename the items in scene to represent what they are testing?

I noticed in the BackfaceCulling scenes, that each object, mesh, and material had a generic name, and made it difficult to figure out what the scene was testing at a glance.

noalak:
Hi Drigax,

Thanks for the quick review.

For the babylon attributes dependency node retreival, we will add an additionnal check on the object ID.
As a side note, the id associated to each MPxNodeClass is currently a debug id. It is adviced to ask Autodesk for a unique id. See documentation on [how to and why](http://mayaid.autodesk.io/). We would need 3 IDs, one for each material type.

For the dryrun requirement, there is a callback when a material is created in 3ds Max, but not in Maya (for once!). For Maya, we can add a callback to the Idle event, triggered every frame when Maya is not busy. We will store a list of materials for which we have already added the babylon attributes node. New materials are detected by comparing this list with the list of all materials. This procedure should not be time consuming and effective even when the number of materials remains unchanged (if a single script remove a material and create a new one).
We also have to make sure the list of babylon materials is reset when a new scene is created or a file is loaded (using dedicated callbacks).
It requires some time for Maya but seems totally feasible.

For the backface culling scene, we do not experiment the scene corruption. We are testing on 2019 though. We will erase the babylon sample file  and create a new one.

Changing the node and material names as well.

Best regards

Drigax:
Thanks for the fast response. I'm hesitant about implementing a per-frame solution in Maya, but since there seems to be little alternative that will give us what we want, I'm willing to move forward with this as a stopgap, and consult Autodesk regarding if they can expose the type of event we want for a future enhancement. 

I am concerned for performance in cases where we have hundreds of materials in HyperShade, so our list checking logic should be as efficient as possible.

Drigax:
I want to amend this previous statement after talking with @PatrickRyanMS. The solution for Max sounds good, we should move forward with this, and instead get the Maya node to initialize on creation and/or connection, so that an end user can pull in an AiStandardSurfaceMaterialNode, manually connect the node in their scene, and be able to modify its parameters without needing to wait for an export pass to actually call Init() for the first time.

SharedProjects/Babylon2GLTF/GLTFExporter.Light.cs
```diff
@@ -30,6 +30,9 @@ private int AddLightExtension(ref GLTF gltf, BabylonLight babylonLight)
                 intensity = babylonLight.intensity,
             };
 
+            // Custom user properties
+            //light.extras = babylonLight.metadata;
```
Selmar: commented on purpose?

noalak: Hi Selmar,

Good point! We commented the extras export for gltf lights as they are not loaded by the gltf loader of Babylon for now.
The export can be done by simply decommenting this line.

For the import, the question is where to perform the convertion of extra to metadata. From a conception point of view, both _ExtraAsMetadata.ts_ and _KHR_lights_punctual.ts_ extensions are required to perform the task. However the extension system doesn't provide a simple way to perform a special task when two extensions are active at the same time.

What we offer to do is copy/paste the _assignExtras method to _KHR_lights_punctual.ts_ file and assign the extras as metadata when loading the node.

@Drigax If it's ok for you we can perform a PR on the babylon repository.

Drigax: @noalak please do, that sounds like a good solution :)

Maya/Samples/glTF 2.0/BabylonMaterialAttributes/MaxSimultaneousLights.gltf
```diff
@@ -0,0 +1 @@
...s":{"baseColorFactor":[0.0,0.0,0.8,1.0],"metallicFactor":0.0,"roughnessFactor":0.1},"emissiveFactor":[0.0,0.0,0.0],"alphaMode":"OPAQUE","name":"aiStandardSurface1"},{"pbrMetallicRoughness":{"baseColorFactor":[0.25,0.25,0.25,1.0],"metallicFactor":0.0,"roughnessFactor":0.450053632},"emissiveFactor":[0.0,0.0,0.0],"alphaMode":"OPAQUE","name":"lambert1"},{"pbrMetallicRoughness":{"baseColorFactor":[0.0,0.0,0.8,1.0],"metallicFactor":0.0,"roughnessFactor":0.1},"emissiveFactor":[0.0,0.0,0.0],"alphaMode":"OPAQUE","name":"pasted__aiStandardSurface1"},{"pbrMetallicRoughness":{"baseColorFactor":[0.5,0.0,0.0,1.0],"metallicFactor":0.0,"roughnessFactor":0.2581639},"emissiveFactor":[0.0,0.0,0.0],"alphaMode":"OPAQUE","name":"pasted__phong1"},{"pbrMetallicRoughness":{"baseColorFactor":[0.0,1.0,0.0,1.0],"metallicFactor":0.0,"roughnessFactor":0.33},"emissiveFactor":[0.0,0.0,0.0],"alphaMode":"OPAQUE","name":"pasted__StingrayPBS1"}],"animations":[{"channels":[{"sampler":0,"target":{"node":11,"path":"rotation"}},{"sampler":1,"target":{"node":11,"path":"translation"}}],"samplers":[{"input":35,"interpolation":"LINEAR","output":36},{"input":37,"interpolation":"LINEAR","output":38}],"name":"All Animations"}]}
```
Selmar: one issue in this file: `"scale":null`

3ds Max/Samples/glTF 2.0/BabylonMaterialAttributes/TransparencyMode/Mask/Mask.gltf
```diff
@@ -0,0 +1 @@
...teStride":12,"name":"bufferViewFloatVec3"},{"buffer":0,"byteOffset":224640,"byteLength":69120,"byteStride":8,"name":"bufferViewFloatVec2"}],"buffers":[{"uri":"Mask.bin","byteLength":293760}],"materials":[{"pbrMetallicRoughness":{"baseColorFactor":[1.0,1.0,1.0,1.0],"baseColorTexture":{"index":0,"texCoord":0},"metallicFactor":0.0,"roughnessFactor":0.450053632},"emissiveFactor":[0.0,0.0,0.0],"alphaMode":"MASK","name":"RedMat","extras":{}},{"pbrMetallicRoughness":{"baseColorFactor":[0.0,1.0,0.0,1.0],"baseColorTexture":{"index":1,"texCoord":0},"metallicFactor":0.0,"roughnessFactor":0.0},"emissiveFactor":[0.0,0.0,0.0],"alphaMode":"MASK","name":"Material #28"},{"pbrMetallicRoughness":{"baseColorFactor":[1.0,1.0,1.0,1.0],"baseColorTexture":{"index":2,"texCoord":0},"metallicFactor":0.0,"roughnessFactor":0.2},"emissiveFactor":[0.0,0.0,0.0],"alphaMode":"MASK","name":"Material #29"}],"textures":[{"sampler":0,"source":0,"name":"RedMat_baseColor.png"},{"sampler":0,"source":1,"name":"02550.png"},{"sampler":0,"source":2,"name":"00204.png"}],"images":[{"uri":"RedMat_baseColor.png"},{"uri":"02550.png"},{"uri":"00204.png"}],"samplers":[{"magFilter":9729,"minFilter":9987,"wrapS":10497,"wrapT":10497}]}
```
Selmar: "extras":{}   bloating the data

Drigax:
Hey @noalak, what's your status regarding these revisions?

noalak:
Hi Drigax, we are progressing well on the revisions.

Done:
- babylon material node retreival for Maya using MPxNodeClass IDs
- add a callback for 3ds Max when a material is created. When triggered it adds the babylon material attributes immediately
- recreate backface culling sample scene
- rename all nodes and materials from sample files
- export extras for glTF lights and import them as metadata in babylon
- support double sided material for Maya

ToDo:
- add a callback for Maya when a material is associated to a mesh. When triggered it adds the babylon material node immediately and connects it to the native material
- null scale for camera
- clean unnecessary empty extras

Next Monday is a day off in France. We are planning to push an update on Tuesday.
A new PR on Babylon.js repository will be linked to this one (for glTF lights extras).
All remaining fixes will be included, except for the Maya callback (expected for middle of next week).

Maya/Samples/glTF 2.0/BabylonMaterialAttributes/MaxSimultaneousLights.gltf
Drigax: I think I already addressed this issue via #645, @noalak, this should be resolved after merging master

Drigax:
Thanks for the update. I think null scale for camera is handled already via #645, but feel free to confirm. 

Regarding Maya material callback, we should also handle the case where the material is not associated with an object, but the user manually adds an babylonXXXNode to the scene, and connects it to a material, (is that driven via MNodeMessage::addAttributeChangedCallback?) we should be able to have the material initialize then, rather than only when a mesh is assigned to use the material.

noalak:
@Drigax We have updated the PR. The only remaining feature to implement is the Maya callback for babylon material node.

Drigax:
@noalak, do you have an eta for the maya callback?

noalak:
Hi Drigax,

We have spent last week trying to implement a solution for Maya. Unfortunately, Maya is very restrictive on callbacks so we can't offer a proper solution.

Our original idea was to add a callback when a material is associated to a mesh. But to add such callback, we need a reference to the material. The only way to implement such solution is to add a callback on the idle event. When triggered, we could loop over all materials and add the mesh connection callback.

Regarding manual setup, we updated the babylonXXXNode creation process so that when instanciated it contains its appropriate attributes (instead of being empty).
When instancied manually, the babylonXXXNode is automatically connected to a shading engine node which has no use. There is no way (event) to delete this node after the babylonXXXNode is created. So the user can either keep it (visual polution) or delete it manually.

If you are ok with the first solution (using idle event), we can implement it.
Otherwise, we stay we the current two options:
- automatic setup at runtime
- manual setup

Drigax:
Great!
I'll create a drop for me and @PatrickRyanMS to test to see if we can work with the current options.

Drigax:
[Max2Babylon2020-6306982b.zip](https://github.com/BabylonJS/Exporters/files/3871036/Max2Babylon2020-6306982b.zip)
Here are the Max2020 bins (I'm assuming that the templates haven't changed since our original iteration):

and here are the Maya2019 bins:
[Maya2Babylon2020-6306982b.zip](https://github.com/BabylonJS/Exporters/files/3871044/Maya2Babylon2020-6306982b.zip)

Drigax:
Currently I just have 2 minor concerns:
In 3ds Max:

image

the checkboxes of the node work, but the number input doesn't. As a user I cannot change max simultaneous or max simultaneous lights using the node inputs or by modifying the node directly.  Is this a known limitation?

In Maya:
image
we're limited in the number of nodes that we can connect to a given babylon node, likewise the nodes of this material node are not exposed to the node graph and aren't controllable, is there a way to expose these node values to the graph as well?

noalak:
For 3dsMax, we haven't found a way to edit or control the attributes displayed on the node in the graph (an equivalent to Maya templates). The huge numeric value for floats is a known issue we have no solution for. Likewise, it doesn't seem possible to control an enum using a dropdown in the node.

For Maya, the babylonXXXNode is currently designed to be connected to a single material. Do you have an example of a node that would allow as many connections as desired (similar to 3dsMax multimaterial)?
We will look at a way to expose babylon material attributes to connections as you wish. Can you explain though what this feature would accomplish ?

noalak:
Status on the remaining tasks:

- #626 Do not export glTF default values: DONE

- #525 Export animations only: IN PROGRESS
More details in the dedicated issue

- #627 Maya duplicated texture object: IN PROGRESS
@Bloadrick is implementing the same optimization process as the one used in 3dsMax

- #599 Expose Maya to MEL commands : TODO

- Update Installer: TODO

- Update Doc: TODO

Drigax:
@noalak, this was just a musing to see if we can remove as many limitations of the system. I realize that the scene materials are used as inputs to our custom attribute node, which explains why we cannot have arbitrary inputs to the system. The goal was to be able to share custom attribute settings between materials to reduce the manual work needed to configure each material in scene. I don't think that this is necessary or necessarily useful given more thought.

Drigax:
@noalak, were you planning on adding the installer changes to this PR? else I'll check this in at its current state.

noalak:
Yes we can add the installer update for tomorrow and push the other features later.
The AE and NE templates will be static (not dynamically retreived from the repo to stick with last update). Ok ?

Also, do we increment the number version as well ? It's currently 1.5.0.
To 1.6.0 ?

Drigax:
@noalak, no need to update the exporter's version number in code, the CI pipeline updates the version number with the build date. I plan to replace the current version number with "private build" or something similar soon via #695

However, we should update the installer version number to 1.3.0, to reflect the change in functionality for both exporters with the need for template installation.

Jaskar:
Hi @Drigax ,

Installer work done ! - @noalak will add it to his PR.

BUT, I need @sebavan to modify the CI pipeline to add the files from AETemplates and NETemplates folders into the .zip file produced. (please :D)

https://github.com/BabylonJS/Exporters/tree/master/Maya/AETemplates
https://github.com/BabylonJS/Exporters/tree/master/Maya/NETemplates

Drigax:
Not a problem, I'll ipdate the CI pipeline asap

noalak:
PR is ready to be merged.

Issue #685, opened 2019-11-14
[3ds Max] If texture folder set to same as glTF location, image uri is incorrect


PatrickRyanMS:
If you set the texture folder to the same location as your glTF file, the uri that is encoded looks like:

..%5C FolderName %2F imageName.ext

The %5C is breaking the loading of the glTF. There could be a use case to store glTF files and images in sibling folders, so we need to figure out how to path between them correctly. 

To start, we should determine if the path to the glTF and the path to the images folders as set in the exporter are the same and if so, just use the file name as the uri rather than going up one folder and then back down. 

It would also be an improvement to the UX if we either spell out in the export window that the texture folder is optional or we add a checkbox to enable the image folder field. Right now, users can interpret the UI as the texture folder is required.

Drigax:
Hey @elpie89 , I remember you added support for exporting to a separate texture folder, can you help explain the use-case for this?

elpie89:
I'll take a look at this asap

Drigax:
Much appreciated.

PR #688, opened 2019-11-21
image uri fix

elpie89:
This should be ok
I also changed the UI as requested
#685

### DIFF
--- a/3ds Max/Max2Babylon/Forms/ExporterForm.cs
+++ b/3ds Max/Max2Babylon/Forms/ExporterForm.cs
@@ -59,11 +59,11 @@ namespace Max2Babylon
         {
             string storedModelPath = Loader.Core.RootNode.GetStringProperty(ExportParameters.ModelFilePathProperty, string.Empty);
             string absoluteModelPath = Tools.ResolveRelativePath(storedModelPath);
-            txtModelName.MaxPath(absoluteModelPath);
+            txtModelPath.MaxPath(absoluteModelPath);
 
             string storedFolderPath = Loader.Core.RootNode.GetStringProperty(ExportParameters.TextureFolderPathProperty, string.Empty);
             string absoluteTexturesFolderPath = Tools.ResolveRelativePath(storedFolderPath);
-            txtTextureName.MaxPath(absoluteTexturesFolderPath);
+            txtTexturesPath.MaxPath(absoluteTexturesFolderPath);
 
             singleExportItem = new ExportItem(absoluteModelPath);
 
@@ -119,9 +119,9 @@ namespace Max2Babylon
 
         private void butModelBrowse_Click(object sender, EventArgs e)
         {
-            if (!string.IsNullOrEmpty(txtModelName.Text))
+            if (!string.IsNullOrEmpty(txtModelPath.Text))
             {
-                string intialDirectory = Path.GetDirectoryName(txtModelName.Text);
+                string intialDirectory = Path.GetDirectoryName(txtModelPath.Text);
 
                 if (!Directory.Exists(intialDirectory))
                 {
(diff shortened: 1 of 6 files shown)


Drigax:
Thanks for the fix, I'll take a look at this first thing tomorrow!

Drigax:
Alright, looks good for the most part, I'd just like to add a warning to our output window when we are placing textures in a directory that's not in a subdirectory of the glTF

3ds Max/Max2Babylon/Forms/ExporterForm.cs
```diff
@@ -138,25 +138,30 @@ private void butModelBrowse_Click(object sender, EventArgs e)
 
             if (saveFileDialog.ShowDialog(this) == DialogResult.OK)
             {
-                txtModelName.MaxPath(saveFileDialog.FileName);
+                txtModelPath.MaxPath(saveFileDialog.FileName);
+                if (string.IsNullOrWhiteSpace(txtTexturesPath.Text))
+                {
+                    string defaultTexturesDir = Path.GetDirectoryName(txtModelPath.Text);
+                    txtTexturesPath.MaxPath(defaultTexturesDir);
+                }
```
Drigax: Can we add a second check here along the lines of:
```

                if (!PathUtilities.IsBelowPath(txtTexturesPath.Text, txtModelPath.Text))
                {
                    {
                        CreateWarningMessage("WARNING: textures path should be below model file path, not all client renderer support this feature",0);
                    }
                }
```

3ds Max/Max2Babylon/Forms/ExporterForm.cs
```diff
@@ -175,14 +180,14 @@ private void btnTextureBrowse_Click(object sender, EventArgs e)
             if (dialog.ShowDialog() == CommonFileDialogResult.Ok)
             {
                 string selectedFolderPath = dialog.FileName;
-                string absoluteModelPath = txtModelName.Text;
+                string absoluteModelPath = txtModelPath.Text;
 
                 if (!PathUtilities.IsBelowPath(selectedFolderPath, absoluteModelPath))
                 {
-                    MessageBox.Show("WARNING: folderPath should be below model file path");
+                    CreateWarningMessage("WARNING: textures path should be below model file path, not all client renderer support this feature",0);
```
Drigax: NIT: can you change "client renderer" to "client renderers"?

SharedProjects/Babylon2GLTF/GLTFExporter.Texture.cs
```diff
@@ -160,7 +160,16 @@ private GLTFTextureInfo ExportTexture(BabylonTexture babylonTexture, GLTF gltf,
...
-                        textureUri = PathUtilities.GetRelativePath( exportParameters.outputPath,exportParameters.textureFolder) + "/"+ name;
+                        textureUri = PathUtilities.GetRelativePath( exportParameters.outputPath,exportParameters.textureFolder);
+                        if (textureUri == ".\\")
+                        {
+                            //texture is in same location of model
+                            textureUri = name;
+                        }
+                        else
+                        {
+                            textureUri = Path.Combine(textureUri, name);
```
Drigax: On windows, we end up having backslashes in our image URI path, after getting this relative path. should we replace that with forward slashes?

SharedProjects/Babylon2GLTF/GLTFExporter.cs
```diff
@@ -482,7 +482,7 @@ private List<GLTFBufferView> SwitchImagesFromUriToBinary(GLTF gltf)
 
             foreach (GLTFImage gltfImage in gltf.ImagesList)
             {
-                var path = Path.Combine(gltf.OutputFolder, Uri.UnescapeDataString(gltfImage.uri));
+                var path = Path.Combine(gltf.OutputFolder, gltfImage.uri);
```
Drigax: Why are we no longer escaping the symbols in our path?

elpie89: it is already unescaped in PathGetRelativePath..this was a repetition.
the gltfimage uri is set only in L176

elpie89:
Everything is now managed in GetRelativePath() method
the relative part is calculated using uring and I removed the part where forward slashes were transformed in backslashes
the Path.Combine() does not alter the slash..so everything is now forwarded.
I think we should stay with URI format, as windows accept both.

Also, I do not remove the first "./"
It is useless and the URI format accept it

the second warning has been added too, in case a model path is set after the textures one

we end up with this 3 cases:

```
"uri": "../../texture/checker.jpg" (with warning raised)
"uri": "./checker.jpg"
"uri": "texture/checker.jpg"
```

If I did not forget some case, should be ok

SharedProjects/Babylon2GLTF/GLTFExporter.cs
```diff
@@ -470,7 +470,11 @@ private string gltfToJson(GLTF gltf)
             // Use the bounded writer in case some values are infinity ()
             using (var jsonWriter = new JsonTextWriterBounded(sw))
             {
+#if DEBUG
```
Drigax: Great change!

PR #690, opened 2019-11-22
load setting on file open, fix behaviour with exporter instance open

elpie89:
when an exporter instance is opened, and the user change the scene, the exporter now loads settings of the scene.
Previously user were forced to close and run an new instance of the exporter


Drigax:
Thanks! I noticed that this was a bit annoying when changing between two scenes that have export settings already saved

3ds Max/Max2Babylon/Exporter/BabylonExporter.cs
```diff
@@ -920,34 +920,34 @@ private void moveFileToOutputDirectory(string sourceFilePath, string targetFileP
...
-                        // The type of node is not exportable (helper, spline, xref...)
-                        break;
-                }
-            }
-            catch (Exception e)
-            {
-                this.RaiseWarning(String.Format("Exception raised during export. Node will be exported as dummy node. \r\nMessage: \r\n{0} \r\n{1}", e.Message, e.InnerException), 2);
-            }
-
+            try
```
Drigax: Can we undo the current branch changes to this chunk? I don't notice a difference other than what's most likely whitespace changes

Drigax:
Looks good. I'll check this in.

PR #699, opened 2019-11-27
Installer version 1.3.1. Support AETemplate and NETemplate subfolder inside published drop.

Drigax:
Update installer version to 1.3.1, add support for installer to no longer exception when processing a directory inside of a ZipFile.


Jaskar:
Hey @Drigax ,

I think that you're talking about the Maya new install with AE and NE files.
These files should be at the root of the plugins folder.

I didn't see it before, but there shouldn't be a "Maya" folder inside the "Maya_2019.zip"

Issue #718, opened 2019-12-13
Maya 2020 Support


BaldSavant:
Hello!

My company is looking to upgrade to Maya 2020 to get some of those sweet new tools, but we've just tried to install the current babylon on it and it doesn't look to be supported yet. I see some threads for specific features for Maya 2020, but nothing about overall support.

If I'm missing something obvious, would you mind pointing me to where I need to go? And if I'm not missing anything, could you tell me when Maya 2020 support is expected?

Thanks!

Drigax:
Thanks for reaching out! Currently we don't support Maya 2020. I'm working closely with @noalak 's team to handle some missing feature support, and I have yet to test the exporter plugin installation as well. 

I'll try to get this up ASAP, but I'll most likely not have this supported by the end of the year.

BaldSavant:
Sure. End of year is no worry, but can I assume based on that response that you'll have support some time fairly early in the new year? As opposed to, say, April and beyond?

Drigax:
Yes, that's definitely the plan. Provided that we don't run into any issues with Maya, we're currently slating this to be done by February

BaldSavant:
Perfect. Thanks for the quick response!

Drigax:
Going to reopen this issue to track progress with Maya 2020 support :)

I expect that we can break this down into 4 subtasks:
- [ ] Create Maya2Babylon project targeting the 2020 SDK
- [ ] Validate proper plug-in functionality
- [ ] Update Installer client for Maya 2020
- [ ] Update Azure CI pipeline to build package and publish the Maya2Babylon 2020 project.

Drigax:
I believe I was able to address task 1 via #729 , however upon loading the plugin into Maya2020, it fails with the following error log:

```
waitCursor -state on;										 $gIgnoreUpdateCallback = true;loadPluginCmd( "C:/Program Files/Autodesk/Maya2020/bin/plug-ins/Maya2Babylon.nll.dll"); 		   updatePluginUI( "92" );waitCursor -state off;									   $gIgnoreUpdateCallback = false;
menu - parent MayaWindow - label "Babylon";
// Result: MayaWindow|menu36 // 
menuItem - label "Babylon File Exporter..." - command "toBabylon";
// Result: MayaWindow|menu36|menuItem940 // 
menuItem - label "Animation groups" - command "AnimationGroups";
// Result: MayaWindow|menu36|menuItem941 // 
// Babylon plug-in initialized // 
[...]

PR #729, opened 2020-01-10
Maya 2020 support

Drigax:
Add Maya2Babylon project targeting Maya2020; Add openmayacs.dll for Maya2020


Drigax:
This PR takes inspiration from #478, I expect that this will only require adding a new project targeting 2020, and adding the 2020 openmayacs library to reference instead of the 2019 library.

I am not going to check this in until I can verify that the built plugin loads properly into the Maya application. Currently, this is blocked by a suspected bug in the maya plugin loader. 

See #718 for more details

Issue #725, opened 2020-01-06
Max2Babylon: instanced bitmap = no transparency?


echadwick-wayfair:
I found some odd behavior when using bitmap instancing in 3ds Max. 
Instancing = wiring a bitmap node output into more than one material input.

![Babylonjs_sandbox](https://user-images.githubusercontent.com/40177036/71849347-d9c94100-309f-11ea-8d49-7699106eaf01.jpg)

![3dsmax_materials](https://user-images.githubusercontent.com/40177036/71849359-de8df500-309f-11ea-8858-ce9ef8e35e70.jpg)

1. If an albedo bitmap is instanced into two materials, one with transparency and one without, then the Alpha channel is discarded upon export.

2. If an albedo bitmap is copied as two inputs, one for each material, then the Alpha channel is properly exported.

This is not a blocker for us, but could prove troublesome later. Also could be confusing for vendors or contractors who export models for us.

[reuse-bitmap-IDs_GLB-files.zip](https://github.com/BabylonJS/Exporters/files/4027857/reuse-bitmap-IDs_GLB-files.zip)

[reuse-bitmap-IDs_3dsMax2020-source-files.zip](https://github.com/BabylonJS/Exporters/files/4027858/reuse-bitmap-IDs_3dsMax2020-source-files.zip)

Max2Babylon 20191220.1
3ds Max 2020.3
Windows 10

Drigax:
@echadwick-wayfair you have a real knack for finding edge cases in our texture export logic 😉

I think this one is a bit complex, currently we export textures as we encounter them per material, and cache which textures were already exported. I assume that may be contributing to this issue, but I'm not totally sure. I'll have to take a look at this first thing tomorrow, but I fear I may want to refactor the current export logic to use multiple passes to first evaluate the textures to be exported from the scene and if we need to process them, then performing the actual processing may help avoid cases like this.

echadwick-wayfair:
Glad to be of assistance :D

Drigax:
Spent some time over the past 2 days investigating this, currently I'm not sure if there is a clean way to handle this.

Currently at export we have 2 materials referencing the same texture source, but the second material requires some additional processing of the texture to add the alpha channel.

At export, we write both textures to disk, using the base color texture name as the final name. In the instanced scene, we end up overwriting the alpha texture with the original texture, and I believe the opposite is happening in the non-instanced scene.

[...]

PR #732, opened 2020-01-16
Give generated BaseColor + Alpha textures a unique internal ID and filename

Drigax:
Modified BaseColorAlphaTexture generation to give combined textures a unique ID and output name.


Drigax:
proposed fix for #725

Issue #739, opened 2020-01-22
Texture exported with incorrect name in glTF


noalak:
Bug repro: 3ds Max file in New.zip from https://github.com/BabylonJS/Exporters/issues/700
Expected result: The textures are exported and linked to Texture object in glTF using their name.
Current result: The name in Texture object does not match texture's name. The import of the scene into the sandbox is failing due to missing texture file "panel_baseColor.jpg". Image in output folder is named "dashboard.jpg".

Drigax:
Closing via #740

PR #740, opened 2020-01-22
Fix texture name when exporting to glTF

noalak:
Fix to https://github.com/BabylonJS/Exporters/issues/739

Use name specified in parameter, if any, rather than input texture name.


Drigax:
Thanks for the fix

Issue #687, opened 2019-11-21
Maya - Scale factor not being applied


aguaviva:
There seems to be a bug when exporting the gltf model with scale factor.The scale factor is not being applied to the model in the gltf export from **Maya**. Even when changing the scale factor it does not seem to add the scaling root in the exported gltf.

This issue was solved for **Max** but we desperately need it for **Maya**.

Drigax:
Sure, I'm sure the fix is similar, as the exporters have a similar architecture. Let me take a quick look...

Drigax:
Unfortunately everything looks functional as far as I can tell.

I'm testing this on Maya 2019, using Maya2Babylon v20191118.2.

![image](https://user-images.githubusercontent.com/3952729/69394139-fa743d80-0c8f-11ea-8499-77626f552f11.png)

Using a scale factor of 3.141593, It appears that the exported glTF has a proper scaling root node added to the scene:
![image](https://user-images.githubusercontent.com/3952729/69394181-21cb0a80-0c90-11ea-894c-085f41c3d45e.png)

Can you confirm your version of Maya and the version of Maya2Babylon that you're using? The more info the better, as it will help me best understand what may be happening here 😊

serendipidipity:
We're still having problems on 20191118.2. I dont see the scaling applied.
We solved this by manually adding scaling nodes to the scene, but it wont scale the light matrices correctly then.

![grafik](https://user-images.githubusercontent.com/20571774/69414534-3d2b2980-0d13-11ea-98fe-d3c024ecbb9d.png)

serendipidipity:
In this case, we need to scale the light matrices at runtime by a factor of 100 (1/0.01) for them to be correct.

Drigax:
Strange...I'm looking into this now, but I can't quite figure out why this is happening:

We check this scale factor here, and this code seems pretty reliable:
https://github.com/BabylonJS/Exporters/blob/cb62e1a30443511db7a0f60982563d597eccab74/Maya/Exporter/BabylonExporter.cs#L321-L331


and testing with your scale factor of 0.01 appears to work well for me as well:
![image](https://user-images.githubusercontent.com/3952729/69467051-492ae180-0d3b-11ea-988a-5b6541b4af48.png)

Drigax:
[...]

PR #745, opened 2020-02-04
Update support for unit scaling in 3dsMax and add it to Maya

### LINKED ISSUE #694: [Maya] Add support for unit scaling
Similarly to #481, we should support scene unit scales natively in the Maya exporter.

.Babylon and glTF expect a default unit of meters, so we should take the current unit used in a Maya scene, and convert to Meters during export.

Related to #687, as our global scene scale is used as a workaround for a case where we always export the scene units as Maya's default of centimeters.

noalak:
Fix #694 for 3ds Max
- Add support for unit scaling when exporting to babylon. Previously only glTF export was supported.
- Fix unit scaling when exporting to glTF.

In 3dsMax, when switching unit system, **all objects in the scene are scaled accordingly**.
Ex:
1 generic unit in a _meter_ unit system is equal to 1m
After switching unit system to _centimeter_, the same 1 generic unit is equal to 1cm

---

Fix #694 for Maya and #687
- Add support for unit scaling to export at meter scale (instead of centimeter as it used to)

In Maya, when switching unit system, **objects in the scene are unaffected**.
Ex:
1 generic unit in a _meter_ unit system is equal to 1m
After switching unit system to _centimeter_, the same 1 generic unit is equal to 100cm

Thus, the support for unit scaling in Maya is straight forward.
All values retreived from Maya API are in centimeter.
The exporter is converting those values to meter.
The current unit system in Maya does not matter during export process.

Also, note that [Maya documentation](https://knowledge.autodesk.com/support/maya/learn-explore/caas/CloudHelp/cloudhelp/2019/ENU/Maya-DataExchange/files/GUID-DAB74E5F-58D2-49E7-B919-BED413F44FDE-htm.html) discourages to modify the unit system:

> You can change the default working units in Maya, which are in metric, but it is recommended to leave them unchanged as some tools in Maya have a set type, for example, Nucleus and Bifrost are hard-coded so that 1 unit = 1 meter, regardless of working units. Instead, **do not change Maya's default working units** and adjust your grid to align with the scale of the project you are working on.

The exporter will work fine regardless of current unit system but it's not the case for all plugins.


Drigax:
Thanks for the PR! Can you please split this PR by issue addressed? That way it's easier to correlate code changes to a specific issue. 

Also, can you please update the samples in a separate PR?

noalak:
All right!

noalak:
PR split in 4: #747, #748, #749 and #750

PR #759, opened 2020-02-13
Line endings fix

elpie89:
Hi, the project is full of mixed  line endings 
I chose to convert all in Linux line endings as most of the files were in this format
I also added in the root of the project a simple power shell utility to convert all line endings 
Also, I added a .gitattributes files to override all user git settings and make sure we keep consistent all line endings in the future
This should free us some future headaches


Drigax:
Thabks for looking into this, much appreciated!

PR #760, opened 2020-02-14
Generic improvements

elpie89:
after the holidays periods and after animation refactoring
to avoid too many problems in our project I made some fixed/improvements only on our fork
now that I have some free time I would like to share back those fix/improvements

1. added some missed parameter initialization in maxscriptmanager..I also added a couple of function
2. the container logic has been refactored to add an incremental ID, this is a case of multiple inheritances of the same container is cleaner and more useful
3. a global variable "BabylonExporterStatus" is now defined during export steps, this is the case of a batch system allow maxscript to eventually know when a export is finished and when to start a new one
4. fixed a problem on uv relative to clamp...to verify export a plane and clamp uv on Y...before the fix ,the result was broken both in babylon and GLTF format...I honestly do not understand ho we didn't find this before
5. relative paths to project folder are now working also for multi export 
6. BabylonBone and BabylonNode are now partial classes to be able to easily extend it in our internal extension


3ds Max/Max2Babylon/Exporter/BabylonExporter.cs
```diff
@@ -165,7 +173,7 @@ public void BakeAnimationsFrame(IINode node,BakeAnimationType bakeAnimationType)
             ScriptsUtilities.ExecuteMaxScriptCommand(@"
                 for obj in selection do 
                 (
-                    if obj.isAnimated == false then continue
+                    --if obj.isAnimated == false then continue
```
Drigax: this is commenting it out, right? Should we just remove this line?

PR #763, opened 2020-02-14
Add ShouldSerialize checks to glTF entities to prevent default value serialization

### LINKED ISSUE #626: Maya glTF exporter writes unnecessary default values
Probably happens for Max also.

For example, https://github.com/bghgary/glTF-Sample-Models/blob/flighthelmet-fix/2.0/FlightHelmet/glTF/FlightHelmet.gltf#L24-L27 (translation is all zeros by default)

These values do not need to be present in the glTF as they are the defaults. It makes the glTF harder to read and bloats the JSON for no reason. This fix should be applied to all values that are written to the glTF.

Here are some other examples where default values should not be written:
* https://github.com/bghgary/glTF-Sample-Models/blob/flighthelmet-fix/2.0/FlightHelmet/glTF/FlightHelmet.gltf#L29-L33 (rotation is identity by default)
* https://github.com/bghgary/glTF-Sample-Models/blob/flighthelmet-fix/2.0/FlightHelmet/glTF/FlightHelmet.gltf#L35-L38 (scale is identity by default)
* https://github.com/bghgary/glTF-Sample-Models/blob/flighthelmet-fix/2.0/FlightHelmet/glTF/FlightHelmet.gltf#L173 (primitive mode is 4 by default)
* https://github.com/bghgary/glTF-Sample-Models/blob/flighthelmet-fix/2.0/FlightHelmet/glTF/FlightHelmet.gltf#L595-L600 (baseColorFactor is all ones by default)
* https://github.com/bghgary/glTF-Sample-Models/blob/flighthelmet-fix/2.0/FlightHelmet/glTF/FlightHelmet.gltf#L603 (texCoord is 0 by default)
* https://github.com/bghgary/glTF-Sample-Models/blob/flighthelmet-fix/2.0/FlightHelmet/glTF/FlightHelmet.gltf#L620-L623 (emissiveFactor is zeros by default)
* etc.

belfortk:
Fixes #626


Drigax:
Don't sell yourself short, you fixed this for both exporters :)

PR #779, opened 2020-03-05
Remove Extra Serialized Bits

### LINKED ISSUE #777: Extra bits are exported to a glTF that shouldn't be
- [x] There is always a "Default light" node that is exported even when not exporting lights.
- [x] `scenes.extensions` has an empty object
- [ ] nodes.extras.currentUVSet is sometimes written out

Repro:
1. Open https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/sourceModels/FlightHelmet/MayaProject/scenes/FlightHelmet.ma in Maya 2019.
2. Export to glTF using Babylon exporter.

The result is this (only the gltf file was included).
[FlightHelmet.zip](https://github.com/BabylonJS/Exporters/files/4278763/FlightHelmet.zip)

belfortk:
removes 'Default Light' always being exported to gltf. removed serialization of empty scenes.extensions and scenes.extras object

Fixes #777


Drigax:
Looks good, thanks!

PR #782, opened 2020-03-09
fix serialization errror with animation group

elpie89:
on a scene with 3000 animation groups and thousand of nodes opening the animation panel was very slow
all data are now filling a dictionary opening the panel, avoiding to retrieve the information over and over from the root node
On the scene I was testing the open animation group panel command is now taking 1-2 second instead of 3-4 minutes


Drigax:
@elpie89, can you please resolve the merge?

elpie89:
> @elpie89, can you please resolve the merge?

done

Issue #781, opened 2020-03-05
[3ds Max 2019] .gltf Transparency not working


dgnsm:
New to web in general so may be missing something obvious here but I cant seem to get opacity/transparency exporting correctly, I am using the stock example 3DS scene files so everything should be set up correctly and I have tried a combination of different export settings. 

The example .GLTF file is loading fine, but I cant figure out the difference between the existing file and the one that I have exported myself. Any help would be great!

I am using 3ds max 2019

log:

```
Exportation started: ***\TransparencyMode_max.gltf
Exporting nodes
	Box001
	Box002
	Pyramid001
	Pyramid002
	Pyramid003
	Pyramid004
	Plane001
	Pyramid005
	Box003
	Box004
	Plane002
	Pyramid006
	Pyramid007
	Pyramid008
	Pyramid009
	Pyramid010
	Total meshes: 16
	Update camera rotation and position
	Update light rotation for glTF export
	No camera defined
	No light defined
	A default hemispheric light was added for your convenience
Exporting materials
	Standard_Color_Opaque
	Standard_ColorMap_Opaque
	Standard_Color_Blend
	Standard_Color_AlphaMap_Blend
	Standard_ColorMap_Blend
	Standard_ColorAlphaMap_Blend
	Wood
	Standard_ColorMap_AlphaMap_Blend
	PBR_Color_Opaque
		Texture type is not supported. Use a Bitmap or a RGB Multiply instead.
		ExportExtraAttributes
	PBR_ColorMap_Opaque
	PBR_Color_Blend
		Texture type is not supported. Use a Bitmap or a RGB Multiply instead.
		ExportExtraAttributes
	PBR_Color_AlphaMap_Blend
		Texture type is not supported. Use a Bitmap or a RGB Multiply instead.
		ExportExtraAttributes
	PBR_ColorMap_Blend
	PBR_ColorAlphaMap_Blend
	PBR_ColorMap_AlphaMap_Blend
	Total: 15
Export animation groups
GLTFExporter | Exportation started
GLTFExporter | Exporting nodes
GLTFExporter | Exporting meshes
GLTFExporter | Exporting skins, lights and cameras
GLTFExporter | Exporting materials
	GLTFExporter.Material | Export material named: Standard_Color_Opaque
	GLTFExporter.Material | Export material named: Standard_ColorMap_Opaque
		GLTFExporter.Material | create gltfMaterial
		GLTFExporter.Material | create gltfPbrMetallicRoughness
[...]

PR #785, opened 2020-03-12
Fix max transparency mode export

Drigax:
Addressing #781


SharedProjects/Babylon2GLTF/GLTFExporter.Material.cs
```diff
@@ -290,7 +293,7 @@ private void ExportMaterial(BabylonMaterial babylonMaterial, GLTF gltf)
                                 opacityBitmap = TextureUtilities.LoadTexture(babylonStandardMaterial.opacityTexture.originalPath, logger);
                             }
 
-                            if (diffuseBitmap != null || specularBitmap != null || opacityBitmap != null)
+                            if ((diffuseBitmap != null && (specularBitmap != null || opacityBitmap != null)) || specularBitmap != null || opacityBitmap != null)
```
Drigax: This change was made to avoid having to generate a new texture if there are no other diffuse textures to pack into it.

The cases are:

There is a diffuse bitmap and a specular or opacity bitmap to be packed into the resultant base color bitmap.

There is no diffuse bitmap, but there is a specular or opacity bitmap to be packed into the base color bitmap.

SharedProjects/Babylon2GLTF/GLTFExporter.Material.cs
```diff
@@ -300,7 +303,14 @@ private void ExportMaterial(BabylonMaterial babylonMaterial, GLTF gltf)
                                 {
                                     logger.RaiseError("Diffuse, specular and opacity maps should have same dimensions", 2);
                                 }
-
+                                baseColorBitmapName = (babylonStandardMaterial.diffuseTexture != null ? Path.GetFileNameWithoutExtension(babylonStandardMaterial.diffuseTexture.name) : TextureUtilities.ColorToStringName(Color.FromArgb(255, (int)(babylonStandardMaterial.diffuse[0]), (int)(babylonStandardMaterial.diffuse[1]), (int)(babylonStandardMaterial.diffuse[2]))))
```
Drigax: This change was made to avoid situations where a generated packed texture shares a name with an unmodified texture in use in the scene. Every packed texture should have a unique name, given unique component textures/colors.

Issue #793, opened 2020-04-06
MAYA babylon exporter texture duplication


kaisunc:
Kind of a workflow problem,

Object with lambert connected to both mayaSG and babylonStandardMaterial will do the following,
1. Exporting with copy texture to output de-selected results in no textures in output folder. (I'm expecting an optimized png)
2. Exporting with copy texture to output selected results in TWO textures in output folder, the source texture, and the optimized png. (Expected).

Object with no babylonStandardMaterial Connected(ie exporting for the first time)
3. Exporting with copy texture to output, but NO babylonSG will result in optimized texture in output folder. BUT babylonSG settings will be set at default. Meaning if I want to set Transparency Mode to something else, I have to run it again, resulting in scenario 1.

Scenario one should result in correct texture in the output folder with no source file included in output folder. Each subsequent export will also retain settings on babylonStandardMaterial like transparency modes.

Drigax:
The expectation for 1) doesn't seem to follow. You expect an output PNG when we disable copying textures? Also what do you mean by "optimized PNG"?

kaisunc:
![export1](https://user-images.githubusercontent.com/8186855/78630132-80a1c100-78cb-11ea-82ea-9c1520d1ec9c.png)
![export2](https://user-images.githubusercontent.com/8186855/78630143-8697a200-78cb-11ea-89a6-6f450e8cb202.png)
![export3](https://user-images.githubusercontent.com/8186855/78630146-88616580-78cb-11ea-8240-11eff8129d45.png)

Scenario1: default settings where copy texture to output is enabled. BabylonStandardMaterialNode is connected manually and transparency mode is set to cutoff. Two texture files are outputed, one with the original texture name, this one is using the copyImage functio. The second texture name is from shader + basecolor, this is the one used in the babylon file. This one is using the saveImage function. My user using the Max exporter is actually expecting a single image but using the original image name, not a generated name. So this needs to be cleared up, 1. saveImage or copyImage? 2. Rename or use source name?  

The third scenario would fit our needs except that transparency modes can't be set as the node is automatically created and attached before exporting. And the second time around would end up as scenario 1. So this is really a moot point.

The second scenario I also wonder if the babylon file will use the original texture path, but it doesn't. If i don't want to copy texture to output, shouldn't it just use the original texture path? Thus, textures can keep on updating?

Drigax:
Can you please upload a sample scene? I don't want to make a guess as to your configuration.

[...]

PR #805, opened 2020-04-15
[Maya] Fix texture duplication when shared opacity and diffuse maps are used.

Drigax:
fix for issue where textures are duplicated if a shared diffuse and opacity map is used.


Drigax:
Fixes #793

Issue #804, opened 2020-04-14
glTF Animation Samplers exporting as "0" instead of "LINEAR"


stanlo:
https://github.com/BabylonJS/Exporters/blob/afe60bd3fd4be73f28f4cbe16b439644405a472d/SharedProjects/GltfExport.Entities/GLTFAnimationSampler.cs#L24

It appears to be exporting the value as an integer, when the glTF spec requires it to be a string.

Additionally, the default is linear, so ShouldSerializeinterpolation appears to be backwards and should export only if **not** linear.

Drigax:
Yikes, that doesn't sound right at all..im guessing we have an enum for the sampler value, but we're not serializing the value as the stringified enum entry.

Mind making a PG to demonstrate this?

stanlo:
Sorry, what is a PG? I can provide a resultant glb. Here's the animated cube from the gltf samples:
```

  "animations": [
    {
      "channels": [
        {
          "sampler": 0,
          "target": {
            "node": 0,
            "path": "rotation"
          }
        }
      ],
      "samplers": [
        {
          "input": 5,
          "interpolation": 0,
          "output": 6
        }
      ],
      "name": "All Animations"
    }
  ]
```

Drigax:
Sorry, got the repos mixed up. We use Playgrounds in the Babylon.JS repo as a way to quickly reproduce an issue.

Usually its easiest to debug an issue when the source Maya or 3dsMax scene is included in the issue as an attachment. I may have a simple scene I can create to reproduce this, but none of the sample scenes appear to be available in a native format for Maya or 3dsMax, so it takes additional time on my side to look into reported issues, as I have to guess as to what you're doing to see this issue.

PR #806, opened 2020-04-15
Fix gltf animation interpolation mode serialization

Drigax:
Fix for gltf animation sampler serialization, changed serialization to use assigned string value of enum, and added attribute for string conversion.git status


Drigax:
Fixes #804

Issue #807, opened 2020-04-17
The GLTF export for Maya will apply animations to nodes that have none applied to them when selecting "Export Animations", even when there are no animations at all.


imbrig:
1. Install the latest BabylonJS GLTF exporter for Maya. Currently the exported gltf says "babylon.js glTF exporter for Maya 2018 v20200415.4"
2. Load up the attached Maya 2019 project that has no animations.
3. Making sure that there are no animations on any nodes, select Babylon export.
4. Enable the "Export Animations" and export the gltf.
5. Open the GLTF file in VSCode and observe the animations on the nodes.
 *Alternatively:
3. Record animations on a specific node or load up the included scene with animations which only has rotation animation on the head joint.
4. Select Babylon export and enable the "Export Animations" and export the gltf.
5. Open the GLTF file in VSCode and observe the animations on the nodes.

Observed Results:
Animations are applied to all of the nodes no matter if animations where recorded on them or not in Maya.

Expected Results:
It is expected only nodes with applied animations in Maya to be the only ones to be exported with animations.

Note: This has been tested on skinned assets and the files used have a skinned mesh. I'm guessing that this might also be an issue with meshes that are not skinned but easier to manage and work around. Also that deleting the animations by hand on the nodes that should not have any, does not affect the gltf negatively and will still load and work in any gltf viewer.

In the Maya zip file there is a project with 2 scenes, one has no animations and the other has animations on the Head joint, recorded in only rotations.
[Maya_Mannequin.zip](https://github.com/BabylonJS/Exporters/files/4495232/Maya_Mannequin.zip)
The GLTF zip has both exported files from the above projects which where exported with the  "Export Animations" checkbox on.
[GLTF_Mannequin.zip](https://github.com/BabylonJS/Exporters/files/4495233/GLTF_Mannequin.zip)

Drigax:
Thanks for the sample, I'll see where we may be exporting empty animations.

Drigax:
Ok, I think i understand what's going on. Currently we bake bone animations frame by frame on export, and after the bake we only include the animations that actually "do" something vua IsAnimationKeysRelevant. Only problem is that we're not checking for animation keyframes that would export as equivalent to the TRS of the node the bone is mapped to, so we currently include them (and any texture transform animations for some reason) and give them two identical keyframes. not good.

I should have a fix for this quickly.

imbrig:
[...]

PR #809, opened 2020-04-23
Fix animations being exported with useless tracks

Drigax:
Fix for issue where non-animated nodes are exported with useless animation tracks. added extra logic to IsAnimationKeysRelevant to take in a base value (ideally, the exported node value) and compare that against the key values, or a default value if not provided before keys are included in export


Drigax:
Fixes #807

PR #849, opened 2020-06-18
Add version warning when installing maya 2020

Drigax:
this PR adds an installation warning to Maya2020 notifying the user to ensure that Maya2020.1 or later is installed.


Drigax:
Added to help ease transitions due to #718

Issue #861, opened 2020-07-31
Maya To GLTF Missing Mesh


philSG:
If I export with mesh optimisations `on` a mesh get lost in the conversion.

with mesh optimizations

```json
    {
      "skin": 0,
      "mesh": 0,
      "name": "Head"
    },
    {
      "name": "Body"
    },
```

without

```json
    {
      "skin": 0,
      "mesh": 0,
      "name": "Head"
    },
    {
      "skin": 1,
      "mesh": 1,
      "name": "Body"
    },
```

I can't provide the asset unfortunately, but this is in the Maya graph.

![Capture](https://user-images.githubusercontent.com/59562074/89049140-b53cfb00-d348-11ea-8fae-1d0c02a15fa7.PNG)

I realise there are a million things that can go wrong here.

philSG:
So I've got a fix for this that works for us.

```C#
// GlobalVertex.cs
if (UV2 != null && other.UV2 != null && !other.UV2.IsAlmostEqualTo(UV2, Tools.Epsilon))
{
    return false;
}
```

`other.UV2` was `null`, and adding a check solves our issue. Happy to raise a PR for this, but unsure if I should add this change to the other attributes (`Position`,`Normal`,UV`,`Weights`...)?

Drigax:
Yes, thanks for looking into this! I agree that we should be doing  a more sophisticated check here.

Imagine 3 scenarios:

v1 does not have uv2, v2 does not have uv2:
these have equivalent uv2 values.

v1has uv2, v2 does not have uv2, or v1 doesnt have UV2, v2 has UV2:
one vertex has coordinates, the other doesnt, we can't consider these vertices as equivalent.

v1 has UV2, v2 has UV2:
we should consider these as equivalent if they both have equivalent uv2 values.

Using this criteria, I'd rather we do something similar to:

```
if (UV2 == null && other.UV2 != null || UV2 != null && other.UV2 == null || !other.UV2.IsAlmostEqualTo(UV2, Tools.Epsilon)
{
    return false;
}
```

Drigax:
Would you like to open a PR with these changes? :)

philSG:
Yep I can do these changes :+1:

deltakosh:
Hey @philSG do you mind sending a new PR?

philSG:
Yes Sorry I got side tracked with a high priority thing, I'm getting back on this area today.

pandaGaume:
[...]

PR #864, opened 2020-08-13
add null checks

philSG:
https://github.com/BabylonJS/Exporters/issues/861

Some extra checks incase something is null.

Slightly different on what was talked about but it should cover the cases with less code.

### DIFF
--- a/Maya/Exporter/GlobalVertex.cs
+++ b/Maya/Exporter/GlobalVertex.cs
@@ -52,37 +52,37 @@ namespace MayaBabylon
                 return false;
             }
 
-            if (!other.Position.IsAlmostEqualTo(Position, Tools.Epsilon))
+            if ((Position != null || other.Position != null) && !other.Position.IsAlmostEqualTo(Position, Tools.Epsilon))
             {
                 return false;
             }
 
-            if (!other.Normal.IsAlmostEqualTo(Normal, Tools.Epsilon))
+            if ((Normal != null || other.Normal != null) && !other.Normal.IsAlmostEqualTo(Normal, Tools.Epsilon))
             {
                 return false;
             }
 
-            if (UV != null && !other.UV.IsAlmostEqualTo(UV, Tools.Epsilon))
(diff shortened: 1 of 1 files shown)


Maya/Exporter/GlobalVertex.cs
```diff
@@ -52,37 +52,37 @@ public override bool Equals(object obj)
                 return false;
             }
 
-            if (!other.Position.IsAlmostEqualTo(Position, Tools.Epsilon))
+            if ((Position != null || other.Position != null) && !other.Position.IsAlmostEqualTo(Position, Tools.Epsilon))
```
Drigax, line 55: Hey, thanks for the contribution,

I'm not sure if this is logically equivalent to what we talked about.

I was looking for `if ((A.Position == null XOR B.Position ==null) || a.Position !~= B.position)  return false;`

So that if one has the attribute, and the other doesn't, then we can't consider the vertices equivalent.

This change seems to be more like ( if (A or B is not null) && A!=B) return false

Which could work if we did the null checking in `Tools.IsAlmostEqualTo`

But the way that function is implemented, it will cause a runtime exception if we pass it a null value:

https://github.com/BabylonJS/Exporters/blob/356add4f425ce0119f5dbaa1bd5a7a3952993096/Maya/Tools/Tools.cs#L88-L104

deltakosh: Makes sense..I reverted it :)

philSG: oh man yeah sorry you're right! braindead!

philSG: This is what I had in my head.

```c#
if (!(Position != null && other.Position != null) && ...)
```

Assuming its like C++ and doesn't evaluate the second part

deltakosh: It is the case :) Do you mind sending a new PR?

PR #886, opened 2020-11-04
Switch normals based on normals being used

### LINKED ISSUE #885: Switch normals based on normals being used
Normals are switched indiscriminately, rather than the normals being populated.

[This is the offending code](https://github.com/BabylonJS/Exporters/blob/745204baf1e092ca0830d771c62596e017e1fbc7/SharedProjects/Babylon2GLTF/GLTFExporter.Mesh.cs#L72)

PR imminent

JohnLudlow:
fixes #885 

I also included a change which allows us to use some symbol definitions to suppress the use of two libraries which we're not interested in and don't want to ship.

### DIFF
--- a/SharedProjects/Utilities/TextureUtilities.cs
+++ b/SharedProjects/Utilities/TextureUtilities.cs
@@ -55,12 +55,18 @@ namespace Utilities
                 {
                     switch (Path.GetExtension(absolutePath).ToLower())
                     {
+#if USE_GDIMAGE_LIBRARY
+
                         case ".dds":
-                            // External library GDImageLibrary.dll + TQ.Texture.dll
-                            return GDImageLibrary._DDS.LoadImage(absolutePath);
+                                  // External library GDImageLibrary.dll + TQ.Texture.dll
+                                  return GDImageLibrary._DDS.LoadImage(absolutePath);
+#endif
+#if USE_PALOMA_TARGAIMAGE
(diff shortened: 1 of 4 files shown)


SharedProjects/Utilities/TextureUtilities.cs
```diff
@@ -231,7 +237,8 @@ private static void _convertToBitmapAndSave(string sourcePath, string destPath,
             Bitmap bitmap;
             switch (imageFormat)
             {
-                case "dds":
+#if !DONT_USE_GDIMAGE_LIBRARY
```
Drigax, line 240: i see we have multiple preprocessor defines for USE_GDI_IMAGE_LIBRARY and the opposite DONT_USE_GDI_IMAGE_LIBRARY. Is this necessary? Can we just use a single define for whether or not to use this library?

JohnLudlow: Oops

I'd started with USE_GDI_IMAGE_LIBRARY and changed it to DONT_USE_GDI_IMAGE_LIBRARY to avoid breaking existing code. Just didn't change everything

JohnLudlow: @Drigax I've updated the PR

Drigax (approved):
Looks good, Thanks

PR #887, opened 2020-11-05
Texture Formats

### LINKED ISSUE #844: 3ds Max 2021: GLB export merges O/R/M jpg files into a ORM png file
I exported a glb file and unpacked it to gltf (via Visual Studio Code gltf tools); even though all source files are jpg files, the mixmaps are exported as png:

image

This does NOT happen when exporting to gltf; here the mixmaps are generated as jpg

pandaGaume:
fix #844 Optimize format for ORM. fix size difference between GLTF and GLB. Replace raw format on GLB with corresponding Encoded format such jpg or png


Drigax (commented):
These changes look pretty good, I just want to make sure we're properly documenting the behavior of the given texture export policies to minimize confusion.

Drigax (approved):

PR #890, opened 2020-11-07
Allow Morph to be exported within animation group 

### LINKED ISSUE #889: Morph are not exported when using animation group
Morph animation (channels and samplers under animation node) are not exported when using animation group. Code path is actually different when dealing with animation group and without.
[moph.zip](https://github.com/BabylonJS/Exporters/files/5504697/moph.zip)

pandaGaume:
Allow morph to be exported within Animation group fix #889
Make the AnimationGroup export and Animation export using the same code path.


Drigax (approved):
Looks like a pretty reasonable refactor

PR #893, opened 2020-11-18
Export morph target names to GLTF

dkyteasSG:
Add mesh.extras.targetNames to gltf export.
Use "{blendShapeDeformer.name}.{currentWeightName}" as the name for babylonMorphTarget.


Drigax (approved):
Looks good to me, thanks!

Issue #892, opened 2020-11-17
Add new glTF extensions support


thomlucc:
As new gltF are being ratified, support needs to be added in the Exporters:

- [ ] [KHR_materials_clearcoat](https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Khronos/KHR_materials_clearcoat/README.md) 
- [ ] [KHR_materials_sheen](https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Khronos/KHR_materials_sheen/README.md)
- [ ] [KHR_materials_transmission](https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Khronos/KHR_materials_transmission/README.md)

pandaGaume:
ClearCoat and Sheen are available into Babylon PBR Material, but how about transmission wich is only covered by alpha (i guess..)?

pandaGaume:
@bghgary, @thomlucc,  as Max did not support sheen, idea should be to create a Material Plugin, similar to the one used for “games PBR” material such as Metal/Roughness or Specular/Glossiness. Then we introduce our own "Babylon material" to allow new properties and hide others... this should also come with the support of these 2 “games PBR” material. This could be done by script.

bghgary:
This might be good to check with Autodesk. Let's make sure they are in the loop on how to handle this.

pandaGaume:
They're on the loop, I had exchange with Brent, and we've already discussed the different options. This proposal is the one he suggests. We may decide to explore this further. We are trying to setting up a call soon to discuss this topic along others.

bghgary:
Sounds perfect!

pandaGaume:
As discussed, we need before this, to modify the physical material to beeing exported as PBRMaterial instead as PBRMetallicRoughnessMaterial

pandaGaume:
The support is coming with 

- [ ] Millestone 1 : Max2023.
- [ ] Millestone 2 : Support of GltfMaterial.
- [ ] Millestone 3 : Support additional KHR_Extensions for PhysicalMaterial Max
- [ ] Millestone 4 : Support additional KHR_Extensions for PhysicalMaterial Maya

kaihekaihe:
Hi，Is [KHR_materials_sheen] supported in Babylon exporters for maya 2023

PR #894, opened 2020-11-18
KHR extensions

### LINKED ISSUE #872: glTF with KHR_texture_transform contains default values that shouldn't be written
Repro:

1. Export a model with texture transform.
2. Look at the glTF.

Result:
```json
"KHR_texture_transform": {
    "offset": [
        0.0,
        0.0
    ],
    "scale": [
        30.0,
        -30.0
    ],
    "texCoord": 0
}
```

Expected:
```json
"KHR_texture_transform": {
    "scale": [
        30.0,
        -30.0
    ]
}
```

pandaGaume:
Fix #872 default test behavior and prepare #892 support for clearcoat, sheen and transmission

### DIFF
--- a/SharedProjects/GltfExport.Entities/GLTFMaterial.cs
+++ b/SharedProjects/GltfExport.Entities/GLTFMaterial.cs
@@ -8,6 +8,10 @@ namespace GLTFExport.Entities
     [DataContract]
     public class GLTFMaterial : GLTFIndexedChildRootProperty
     {
+        internal static float[] one2 = { 1.0f, 1.0f };
+        internal static float[] zeros2 = { 0.0f, 0.0f };
+        internal static float[] zeros3 = { 0.0f, 0.0f, 0.0f };
+
         public enum AlphaMode
         {
             OPAQUE,
@@ -63,7 +67,7 @@ namespace GLTFExport.Entities
 
         public bool ShouldSerializeemissiveFactor()
         {
-            return (this.emissiveFactor != null && !this.emissiveFactor.SequenceEqual(new float[] { 0f, 0f, 0f }));
+            return (this.emissiveFactor != null && !this.emissiveFactor.SequenceEqual(zeros3));
         }
 
         public bool ShouldSerializealphaMode()
@@ -83,7 +87,7 @@ namespace GLTFExport.Entities
     }
 
 
-    // https://github.com/AltspaceVR/glTF/blob/avr-sampler-offset-tile/extensions/2.0/Khronos/KHR_texture_transform/README.md
+    // https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_texture_transform
     [DataContract]
     public class KHR_texture_transform
     {
@@ -102,7 +106,7 @@ namespace GLTFExport.Entities
 
         public bool ShouldSerializeoffset()
         {
-            return (this.offset != null && this.offset != new float[] { 0f, 0f });
+            return (this.offset != null && !this.offset.SequenceEqual(GLTFMaterial.zeros2));
 
         }
         public bool ShouldSerializerotation()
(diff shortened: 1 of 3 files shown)


SharedProjects/GltfExport.Entities/GLTFMaterial.cs
```diff
@@ -63,7 +67,7 @@ public bool ShouldSerializeemissiveTexture()
 
         public bool ShouldSerializeemissiveFactor()
         {
-            return (this.emissiveFactor != null && !this.emissiveFactor.SequenceEqual(new float[] { 0f, 0f, 0f }));
+            return (this.emissiveFactor != null && !this.emissiveFactor.SequenceEqual(zeros3));
```
Drigax, line 70: I reccomend using tools.isAlmostEqualTo() instead of doing floating point equivalency.

SharedProjects/GltfExport.Entities/GLTFMaterial.cs
```diff
@@ -8,6 +8,10 @@ namespace GLTFExport.Entities
     [DataContract]
     public class GLTFMaterial : GLTFIndexedChildRootProperty
     {
+        internal static float[] one2 = { 1.0f, 1.0f };
```
Drigax, line 11: can we instead add a definition to our Tools.Math file with definitions of vector2, vector3 and the zero and one constants, similar to babylon.js?

Drigax (changes requested):

SharedProjects/GltfExport.Entities/GLTFMaterial.cs
pandaGaume: I agree, Then we use the MathUtilities wich is already used in GLTFLight, then reduce the dependencies. 
and add 
```
        public static bool IsAlmostEqualTo(this float[] array, float value, float epsilon)
        {
            return ! array.Any( v=> Math.Abs(v-value) > epsilon);
        }
```
into the game
ok ?

SharedProjects/GltfExport.Entities/GLTFMaterial.cs
pandaGaume: Yes we may have this , but i recommend to introduce this vector class directly into the GltfExport.Entities to avoid new dependencies. The gltf project is supposed to be standalone and self sufficient (wich its not the case today). So, similar to babylon.js, introduce one Maths.cs file with Vector2 and Vector3, and some logic to replace the MathUtilities used in  only one place so far.

SharedProjects/GltfExport.Entities/GLTFMaterial.cs
Drigax: Sounds good, thanks

SharedProjects/GltfExport.Entities/GLTFMaterial.cs
Drigax: That seems like an alright decision, I just like code reuse, and not making architectural decisions for convenience sake, Currently we take dependencies on some shared utilities, shared projects, as long as we don't take any additional dependencies directly from Max or/Maya2Babylon.

Ultimately, it would be a good goal to architect such that we can make Babylon2GLTF its own independent project.

Contributing to our utilities should still allow the project to be relatively standalone, right? That would be akin to using a shared math library or scientific computing library in my opinion.

Drigax: I definitely agree with adding a helper function to compare arrays via IsAllAlmostEqualTo,

Alternatively you can override IsAlmostEqualTo to use float[] as well.

pandaGaume: Ok this is what i done. I guess PR is ready now.

Drigax:
Thanks, checking in.

Drigax (approved):

PR #898, opened 2020-12-01
modify JSON export encoding to utf8 

renjianqiang:
modify encoding to utf8 when export JSON for .glb format, then the exported glb model can support for languages other than english.


bghgary (approved):
We should make sure to follow the glTF rules here: https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#json-encoding

Drigax (approved):

PR #908, opened 2020-12-07
Maya : rebuild deleted morph target

### LINKED ISSUE #902: Morph not imported when target are deleted on Maya
In order to save space, it's a common  way to delete target for morph, once the design done. when targets are deleted Maya is not storing points position anymore, but only deltas of the affected vertices. 
the result is that [line](https://github.com/BabylonJS/Exporters/blob/45857cb01bf7bd3bf0f4d378a8e99f6b14a61735/Maya/Exporter/BabylonExporter.Mesh.cs#L1267) on which the entire morph export is based on, return null
We MUST find the way to access these delta to rebuild the target at export time.
For the purpose there are several attributes that might be investigate, such "outputGeometry"

pandaGaume:
Add support for deleted Morph target (which is 90% of cases) while target must be, by design and maya documentation, only used at design time. fix #902


Drigax:
Well done! this is a pretty elegant solution to this issue.

Drigax (approved):

PR #910, opened 2020-12-08
Command line build and test support

rdurnin:
Adding support for command line building and testing using mayapy (based on python 3.7.7).
Removed stop play state (MGlobal) command from BabylonExporter.cs as it is unavailable without the UI.
Added create directory call for non-existent export folders to BabylonExporter.cs.
Extending support for Arnold shader types during export (aiImage file type) for cleaner workflows.
Various python and batch files added for cli/cicd support.


Maya/cicd/batch/buildMaya2Babylon.bat
```diff
@@ -0,0 +1,54 @@
+@echo off
```
Drigax, line 1: Hey, I'm not having much luck with this build script, currently the solution fails to build on your branch using this script, but builds in Visual Studio without issue.

Maya/Exporter/BabylonExporter.cs
```diff
@@ -60,13 +60,17 @@ public void Export(ExportParameters exportParameters)
...
+            // The following causes erorrs when trying using mayapy.exe
+            //---------------------------------------------------------------------
             // Check if the animation is running
-            MGlobal.executeCommand("play -q - state", out int isPlayed);
-            if(isPlayed == 1)
-            {
-                RaiseError("Stop the animation before exporting.");
-                return;
-            }
+            // MGlobal.executeCommand("play -q - state", out int isPlayed);
```
Drigax, line 68: should this stay comented? if so, then please remove it instead.

Drigax (changes requested):
Can we make sure that the build scripts are fully functional on a fresh installation?

Maya/cicd/batch/buildMaya2Babylon.bat
rdurnin: Hey, and I have added command line arguments to the batch scripts which will allow different versions of Maya. I do not, however, have any version other than 2021 to test with, and am unsure if versions previous of python will support pytest (etc).
Please try and run the command as follows, and let me know if you continue to have problems:
(using an administrator cmd shell and within the Exporters\Maya directory)
cicd\batch\buildMaya2Babylon.bat MAYAVERSION 
cicd\batch\testMaya2Babylon.bat MAYAVERSION

Maya/Exporter/BabylonExporter.cs
rdurnin: This command appears to be used to stop playback before exporting begins and should probably be moved to AnimationForm.cs where it can be used as part of the windowed (in gui) export routine.

Maya/cicd/batch/buildMaya2Babylon.bat
rdurnin: Make sure you have added the msbuild.exe directory to system %PATH% as well. It is located in the visual studio install directory and is required to build solution files from a cmd shell.

Maya/cicd/batch/buildMaya2Babylon.bat
```diff
@@ -0,0 +1,60 @@
...
+rem  usage:  buildBabylon2Maya
+rem  where:  install dll and nll assemblies to 2021 plug-ins directory:
+rem          located at: C:\Program Files\Autodesk\Maya(Version)\bin\plug-ins
+rem
+
+if "%1" == "" goto nomayaver
+
+set mayaversion=%1
+set mayalocation=%ProgramFiles%\Autodesk\Maya%mayaversion%
+
```
bghgary, line 14: It is generally a good idea to put `setlocal` at the beginning of a batch script to prevent variable leakage.

bghgary: Same for the other batch script.

Maya/cicd/batch/buildMaya2Babylon.bat
```diff
@@ -0,0 +1,60 @@
...
+set installedassembly=%installdir%\Maya2babylon.nll.dll
+if exist "%installedassembly%" del /f "%installedassembly%"
+
+set releasedir=%~dp0..\..\bin\Release\2020
+if exist "%releasedir%" del /f /q "%releasedir%"
+
+set assemblydir=%~dp0..\..\assemblies\2020
+if exist "%assemblydir%" del /f /q "%assemblydir%"
+
+msbuild %solutionfile% /v:quiet /t:Rebuild /p:Configuration=Release
```
bghgary, line 32: It might be good to check for errors after this. Usually goes something like this:
```suggestion
msbuild %solutionfile% /v:quiet /t:Rebuild /p:Configuration=Release
if errorlevel 1 goto :error
```

Same goes for the copying. Not sure if the errorlevel is right for msbuild though. Hopefully, it returns an error code when it fails.

Maya/cicd/batch/buildMaya2Babylon.bat
Drigax: Thanks, I must have had my path setup to use the wrong MSBuild, I also noticed I needed to install `pip` for my mayapy distribution in order to run the test pass.

Drigax (approved):

sebavan:
@rdurnin could you update with master, and I ll merge the branch in :-)

PR #916, opened 2020-12-28
set missing Maya attribute from node

### LINKED ISSUE #913: Maya : Backface culling reset when exporting
Hello,

I'm using the Maya 2019 Exporter 20201214.2 version of the exporter, and I noticed that everytime I export a mesh with the backface culling disabled, it resets. So I'm unable to turn it off and thus unable to have doublesided faces on my meshes.

Here are the parameters I use when exporting. But I doubt it's the problem since I used previous version of the exporter with the same settings and didn't have this problem.
The shader doesn't affect the bug since it happened with lambert and stingray pbs alike.

20201222_BabylonJS-Exporter-settings

I've also linked you the log of the export, if it can help.

Thank you in advance,
Alexis

[20201222_Backface-culling-problem_log.txt](https://github.com/BabylonJS/Exporters/files/5731191/20201222_Backface-culling-problem_log.txt)

pandaGaume:
Maya exporter : Set the babylonBackfaceCulling & babylonUnlit  material bool attribute according the node value
this fix #913


Drigax (approved):

AlexIbanez16:
I have an update concerning the babylonBackfaceCulling problem (resetting the value during the export). It's still there using the latest release of the exporter 20210107.7.

pandaGaume:
Did this happend with a particular material ? I might missed something during my repro and test.

AlexIbanez16:
I only tried with the Stingray PBS. I'll try tomorrow with other materials (lambert, aiStandardSurface, blinn and phong) and keep you updated.

AlexIbanez16:
Okay, so it seems that the problem persists with the Stingray PBS and the aiStandardSurface materials. Maybe it's because they have babylonStingrayPBSMaterialNode and babylonAiStandardSurfaceMaterialNode instead of a babylonStandardMaterialNode.
The lambert, phong, and blinn materials work fine.

pandaGaume:
Ok this is more clear, I migh had corrected only for the standard material and forgot a check somewhere. I will investigate this tomorrow morning my time (which is CET).

AlexIbanez16:
Hi! I'm writing to know if you found what was causing this problem and, if not, can I be of any help?

pandaGaume:
@AlexIbanez16 Sorry for the delay, i was comitted on other works. I have to review what i published because i certainly forgot to correct some other type of material. I will try to do my best to publish a correction asap.

pandaGaume:
@AlexIbanez16  publish a fix. Might be better now. Wait for the PR to be merge.

PR #917, opened 2020-12-28
Max clones

### LINKED ISSUE #835: Feature Request - Export clones
It's possible export clones from 3dsmax to allow instances with different material?

https://doc.babylonjs.com/how_to/optimizing_your_scene

> If sharing the same material is a problem, you can then think about using clones which share the same geometry with mesh.clone("newName")

Like - **Export instances as clones**

Thanks!

pandaGaume:
Add experimental support for Cloning Mesh when instances do not share same material as master. Introduce  Geometries support into Babylon Entities for the purpose. fix #835.


3ds Max/Max2Babylon/Exporter/BabylonExporter.Mesh.cs
```diff
@@ -121,11 +121,15 @@ private BabylonNode ExportMesh(IIGameScene scene, IIGameNode meshNode, BabylonSc
                         // Check if we need to export this instance as an instance mesh.
                         if (tabGuid == masterMeshPair.Key.id)
                         {
-                            // If there is already an exported mesh in the scene that shares this mesh's material, then export it as an instance.
-                            if (masterMeshPair.Key.materialId == null
-                            || (meshNode.NodeMaterial != null && meshNode.NodeMaterial.MaxMaterial.GetGuid().ToString().Equals(masterMeshPair.Value.NodeMaterial.MaxMaterial.GetGuid().ToString())))
+                            bool isShareMat = masterMeshPair.Key.materialId == null || (meshNode.NodeMaterial != null && meshNode.NodeMaterial.MaxMaterial.GetGuid().ToString().Equals(masterMeshPair.Value.NodeMaterial.MaxMaterial.GetGuid().ToString()));
+
+                            BabylonNode n = isShareMat?
+                                ExportInstanceMesh(scene, meshNode, babylonScene, masterMeshPair.Key, masterMeshPair.Value) :
```
Drigax, line 127: The use of a ternary operator makes this code hard to understand at first glance. Can we just use if-else for readability here?

3ds Max/Max2Babylon/Exporter/BabylonExporter.Mesh.cs
```diff
@@ -121,11 +121,15 @@ private BabylonNode ExportMesh(IIGameScene scene, IIGameNode meshNode, BabylonSc
                         // Check if we need to export this instance as an instance mesh.
                         if (tabGuid == masterMeshPair.Key.id)
                         {
-                            // If there is already an exported mesh in the scene that shares this mesh's material, then export it as an instance.
-                            if (masterMeshPair.Key.materialId == null
-                            || (meshNode.NodeMaterial != null && meshNode.NodeMaterial.MaxMaterial.GetGuid().ToString().Equals(masterMeshPair.Value.NodeMaterial.MaxMaterial.GetGuid().ToString())))
+                            bool isShareMat = masterMeshPair.Key.materialId == null || (meshNode.NodeMaterial != null && meshNode.NodeMaterial.MaxMaterial.GetGuid().ToString().Equals(masterMeshPair.Value.NodeMaterial.MaxMaterial.GetGuid().ToString()));
```
Drigax, line 124:
```suggestion
                            bool sharesMaterial = masterMeshPair.Key.materialId == null || (meshNode.NodeMaterial != null && meshNode.NodeMaterial.MaxMaterial.GetGuid().ToString().Equals(masterMeshPair.Value.NodeMaterial.MaxMaterial.GetGuid().ToString()));
```

Drigax: for readibility. Character count is not at a premium here. ;)

3ds Max/Max2Babylon/Exporter/BabylonExporter.Mesh.cs
```diff
@@ -603,29 +607,124 @@ private void ExportUserData(IIGameNode meshNode, BabylonAbstractMesh babylonMesh
...
+                    // ensure to clean the material because Clone() did not define any behavior about
+                    // cloning the material id.
+                    babylonMesh.materialId = null;
+                }
+
+                // Export the custom attributes of this mesh
+                babylonMesh.metadata = ExportExtraAttributes(maxNode, babylonScene);
+
+                // Append userData to extras
+                ExportUserData(maxNode, babylonMesh);
```
Drigax, line 689: Will this result in us having user data from the original mesh, and the clone on the exported clone?

SharedProjects/BabylonExport.Entities/BabylonVertexData.cs
```diff
@@ -0,0 +1,83 @@
+﻿using System.Runtime.Serialization;
+
+namespace BabylonExport.Entities
+{
+
+    [DataContract]
+    public class BabylonVertexData : IBabylonMeshData
```
Drigax, line 7: If the goal here is to abstract the geometry information from the rest of the mesh information, should we remove the geometry-related fields from BabylonMesh, and instead use a facade to the BabylonGeometry when serializing?

SharedProjects/BabylonExport.Entities/BabylonMesh.cs
```diff
@@ -5,12 +5,31 @@
...
+
     [DataContract]
-    public class BabylonMesh : BabylonAbstractMesh
+    public class BabylonMesh : BabylonAbstractMesh, IBabylonMeshData
     {
         [DataMember]
         public string materialId { get; set; }
+        
+        [DataMember]
+        public string geometryId { get; set; }
```
Drigax, line 32: How are we serializing this to glTF?

Drigax (approved):
I understand the sentiment here in how we want to try to share geometry data across meshes that arent full instances of each other, but lets make sure that our architectural changes are appropriate, and not just purpose built, and that we're considering all scenarios.

Drigax (changes requested):
That previous review was supposed to request changes ^^^

SharedProjects/BabylonExport.Entities/BabylonVertexData.cs
pandaGaume: Agree for an SDK, but here i stick with the original design idea which is to have Entities dedicated to serialization. Babylon format, introduce the geometries fields late into version 2 (i guess remember), and for compatibility, the file format is keeping the both. David Catuhe or other Guru  will be able to tell us if this is still in use into Babylon. My concern was, related to our first conversation, was to keep the change "minimal". I will be happy to explore this if you think it's mandatory.

SharedProjects/BabylonExport.Entities/BabylonMesh.cs
pandaGaume: GLTFExporter.Mesh is in charge. 
I choose to abstract the geometry throught an interface 
```
    public interface IBabylonMeshData
    {
        float[] positions { get; set; }
        float[] normals { get; set; }
        float[] tangents { get; set; }
        float[] uvs { get; set; }
        float[] uvs2 { get; set; }
        float[] uvs3 { get; set; }
        float[] uvs4 { get; set; }
        float[] uvs5 { get; set; }
        float[] uvs6 { get; set; }
        float[] colors { get; set; }
        int[] matricesIndices { get; set; }
        float[] matricesWeights { get; set; }
        int[] indices { get; set; }
    }
```
An then it's kind of transparent for the GLTF serializer to work with a mesh data OR geometry. 
This is also partially respond to your previous comment.

3ds Max/Max2Babylon/Exporter/BabylonExporter.Mesh.cs
pandaGaume: yes, data are duplicated

SharedProjects/BabylonExport.Entities/BabylonMesh.cs
pandaGaume: GLTFExporter.Mesh is in charge. 
I choose to abstract the geometry throught an interface 
```
    public interface IBabylonMeshData
    {
        float[] positions { get; set; }
        float[] normals { get; set; }
        float[] tangents { get; set; }
        float[] uvs { get; set; }
        float[] uvs2 { get; set; }
        float[] uvs3 { get; set; }
        float[] uvs4 { get; set; }
        float[] uvs5 { get; set; }
        float[] uvs6 { get; set; }
        float[] colors { get; set; }
        int[] matricesIndices { get; set; }
        float[] matricesWeights { get; set; }
        int[] indices { get; set; }
    }
```
An then it's kind of transparent for the GLTF serializer to work with a mesh data OR geometry. 
This is also partially respond to your previous comment.

3ds Max/Max2Babylon/Exporter/BabylonExporter.Mesh.cs
Drigax: Are we sure thats the intention here? I was under the assumption that user data is instance-specific, I'm not sure that we want to duplicate this data as well.

SharedProjects/BabylonExport.Entities/BabylonMesh.cs
Drigax: I'm not sure if i'm being clear here.

Imagine we have two meshes, one original and a clone of that original mesh in scene:

when serialized both will be serialized as IBabylonMeshData, however the fields in the clone are now null. Currently we are adding no additional logic to the glTF serializer to handle serialization of meshes with GeometryData, so I'm concerned that the second mesh may not be properly serialized, as we will be using those nulled fields.

SharedProjects/BabylonExport.Entities/BabylonVertexData.cs
Drigax: I'm not sure if I'm being clear here.

Imagine we have two meshes, one original and a clone of that original mesh in scene:

when serialized even though we are using the IBabylonMeshData, interface, but we are also not adding any additional logic to the glTF serializer to handle serialization of meshes with GeometryData only, so I'm concerned that the second mesh may not be properly serialized, as we will be using those nulled fields.

Personally I don't believe that we should limit our design for backwards compatibility here. @deltakosh, your input would be appreciated as well.

If we decide to go all in with GeometryData, we should go all in and use this as our storage of geometry, else we're complicating our design with additional usecases which may or may not be valid.
@Drigax

3ds Max/Max2Babylon/Exporter/BabylonExporter.Mesh.cs
pandaGaume: Well, in max, Instance are a "pointer" to a mesh, and then share ALL the values and behavior. Who can clarify the initial intention ?

SharedProjects/BabylonExport.Entities/BabylonVertexData.cs
pandaGaume: I put all (as far i know ) the mandatory logic in place inside the GLTF part of the exporter.  I asked a lot of question inside the issue but no response :-), then i move forward and also provide sample and result at the end of the issue.
I'm agree that this enhancement is quite intrusive and this is why i put it as "experimental". 
So, Let's continue this interesting discussion :-).

Drigax: Cool, missed ALLL of those changes, thanks. Since you're handling both cases relatively elegantly in the glTF exporter, I'm much less worried. Lets check this in and revisit if we run into any issues in the future.

Drigax (approved):
Can you resolve the merge conflicts with the other change to the designer form?

PR #918, opened 2020-12-29
Expose Draco parameters for Max

### LINKED ISSUE #878: Expose settings for Draco
I export a glb file from 3DS Max with BabylonJS tool, the geometry is a bit out of shape after using Draco compression.The shape is totally fine without Draco, but the file size is too big,almost 7 times bigger.
The major problem here is the default property draco.quantizePositionBits in BabylonJS plugin is too low, and there isn’t an option to modify.
I change this to 20, then everything is fine.
The compression level does not affect the geometry at all even set to 10.
BTW, model export from this plugin also got normal problem, the color looks different from FBX file and not good.I’m using FBX now, so I didn’t look into this.

pandaGaume:
Expose Draco compression parameters into the export form of 3DSMax. Fix #878.
We choose to add an additional "Advanced" tab into the form to encompass the Draco Control.

Capture draco

- Draco settings is enabled according to the checkBox "Use Draco Compression".
- Parameters are serialized/deserialized into rootNode attributes.
- Parameters default values are set to Draco official default values
- Reset button get the values QuantizeBits values back to default.

 
Note : We might address the Maya version differently, while there is an issue redesigning the Form. However all the necessary code are located into shared project so the workload will be minimum.


SharedProjects/BabylonExport.Entities/ExportParameters.cs
```diff
@@ -10,6 +10,38 @@ public enum TextureFormatExportPolicy
...
+        public const string dracoPrefix = "draco.";
+        public static readonly string compressionLevel_param_name = $"{dracoPrefix}compressionLevel";
+        public static readonly string quantizePositionBits_param_name = $"{dracoPrefix}quantizePositionBits";
+        public static readonly string quantizeNormalBits_param_name = $"{dracoPrefix}quantizeNormalBits";
+        public static readonly string quantizeTexcoordBits_param_name = $"{dracoPrefix}quantizeTexcoordBits";
+        public static readonly string quantizeColorBits_param_name = $"{dracoPrefix}quantizeColorBits";
+        public static readonly string quantizeGenericBits_param_name = $"{dracoPrefix}quantizeGenericBits";
+        public static readonly string unifiedQuantization_param_name = $"{dracoPrefix}unifiedQuantization";
+
+        public const int compressionLevel_default = 7;
```
Drigax, line 24: How were these default values chosen?

SharedProjects/Utilities/Draco/DracoUserControl.cs
```diff
@@ -0,0 +1,68 @@
+﻿using System;
+using System.Windows.Forms;
+
+namespace Utilities
+{
+    public partial class DracoUserControl : UserControl
+    {
+        public const int quantizePositionBits_default = 14;
```
Drigax, line 8: Same with the values here, do we have some documentation we can refer users to that justifies why we chose these magic numbers?

Drigax (approved):

SharedProjects/BabylonExport.Entities/ExportParameters.cs
pandaGaume: These default values are coming from [GLTF-pipeline](https://github.com/CesiumGS/gltf-pipeline) (the Node tool we use for compression ).

SharedProjects/Utilities/Draco/DracoUserControl.cs
pandaGaume: Same as above. State of art for Draco user :-). But agree, we should put this in comment/documentation

SharedProjects/BabylonExport.Entities/ExportParameters.cs
Drigax: Can we add documentation that points future readers to where we got these values from?

SharedProjects/Utilities/Draco/DracoUserControl.cs
```diff
@@ -0,0 +1,70 @@
+﻿using System;
+using System.Windows.Forms;
+
+namespace Utilities
+{
+    public partial class DracoUserControl : UserControl
+    {
+        // default values are defined from https://github.com/CesiumGS/gltf-pipeline
```
Drigax, line 8:
```suggestion
        // default values are defined from https://github.com/CesiumGS/gltf-pipeline#command-line-flags
```

SharedProjects/BabylonExport.Entities/ExportParameters.cs
```diff
@@ -10,6 +10,40 @@ public enum TextureFormatExportPolicy
...
+        public static readonly string compressionLevel_param_name = $"{dracoPrefix}compressionLevel";
+        public static readonly string quantizePositionBits_param_name = $"{dracoPrefix}quantizePositionBits";
+        public static readonly string quantizeNormalBits_param_name = $"{dracoPrefix}quantizeNormalBits";
+        public static readonly string quantizeTexcoordBits_param_name = $"{dracoPrefix}quantizeTexcoordBits";
+        public static readonly string quantizeColorBits_param_name = $"{dracoPrefix}quantizeColorBits";
+        public static readonly string quantizeGenericBits_param_name = $"{dracoPrefix}quantizeGenericBits";
+        public static readonly string unifiedQuantization_param_name = $"{dracoPrefix}unifiedQuantization";
+
+        
+        // default values are defined from https://github.com/CesiumGS/gltf-pipeline
```
Drigax, line 25:
```suggestion
        // default values are defined from https://github.com/CesiumGS/gltf-pipeline#command-line-flags
```

Drigax (approved):

SharedProjects/Utilities/Draco/DracoUserControl.cs
Drigax: Thank you.

PR #927, opened 2021-01-29
Add support for MaterialId on multiMaterial

### LINKED ISSUE #926: [3ds max] Export does not use the IDs defined on the multi-sub material.
### This is how the material Assignment should look..
3dsmax_viewport_expected look

### This is the material assignment I get on export to glTF
Incorrect material assignment on gltf export

materialNodes_graph
### This is the two multi-sub materials that I have assigned. The first one gets exported as expected but the second one visualizes the problem.
 It looks like that you are assuming that a multi-sub material is just an array and the first item in the multi-sub material is id:1

### When I do maxscripts and work with multi-sub materials I always ask the multi-material to give me its [materialIDList](http://help.autodesk.com/view/3DSMAX/2019/ENU/?query=materialIDList)

That property is exposed in maxscript.. not sure if it is in max.net.
I have two multisub materials.. both containing 4 materials..
Material_ShowingIDs

If you look in the bottom one.. I have set the First material in the multi-sub material to use ID:4
and if I run this in maxscript,with objects using that multimaterial selected,
```
m = $.material
m.materialIDList
```

I get this respons:
```
#Multi/Sub-Object:INCORRECT EXPORT.. LOOK AT THE IDS(Physical Material:RED, Physical Material:YELLOW, Physical Material:GREEN, Physical Material:BLUE)
#(4, 3, 2, 1)

--if I do this on the correctly exported material I get this.
#Multi/Sub-Object:CORRECT LOOKING MULTISUB(Physical Material:BLUE, Physical Material:GREEN, Physical Material:YELLOW, Physical Material:RED)
#(1, 2, 3, 4)
```

So I think there is a bug that you are not reading the actual ID of the material.. you are mainly relying on the order of the material array..

I have attached a 3dsmax file.. that you can test with
[3dsmaxFile_BABYLONJS-MULTIMAT-BUG.zip](https://github.com/BabylonJS/Exporters/files/5888050/3dsmaxFile_BABYLONJS-MULTIMAT-BUG.zip)

pandaGaume:
On Multi-Material, when user set it's own id, we must access the id using Max meshNode.NodeMaterial.GetMaterialID(i) instead of assign sequential index Fix #926

### DIFF
--- a/3ds Max/Max2Babylon/Exporter/BabylonExporter.Mesh.cs
+++ b/3ds Max/Max2Babylon/Exporter/BabylonExporter.Mesh.cs
@@ -808,12 +808,15 @@ namespace Max2Babylon
             int indexInFaceIndexesArray = 0;
             for (int i = 0; i < multiMatsCount; ++i)
             {
-                int materialId = i;
                 var indexCount = 0;
                 var minVertexIndex = int.MaxValue;
                 var maxVertexIndex = int.MinValue;
+                // Material Id is 0 if normal material, and GetMaterialID if multi-material
+                // default is [1,n] increment by 1 but user can decide to change the id (still an int) and set for example [4,3,9,1]
+                // note that GetMaterialID return the user id minus 1
+                int materialId = multiMatsCount == 1? i : meshNode.NodeMaterial.GetMaterialID(i);
                 var subMesh = new BabylonSubMesh { indexStart = indexStart, materialIndex = i };
-
+ 
                 if (multiMatsCount == 1)
                 {
                     for (int j = 0; j < unskinnedMesh.NumberOfFaces; ++j)
@@ -898,10 +901,9 @@ namespace Max2Babylon
                         }
                     }
                 }
-
+ 
                 if (indexCount != 0)
                 {
-
                     subMesh.indexCount = indexCount;
                     subMesh.verticesStart = minVertexIndex;
                     subMesh.verticesCount = maxVertexIndex - minVertexIndex + 1;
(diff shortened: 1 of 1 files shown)


3ds Max/Max2Babylon/Exporter/BabylonExporter.Mesh.cs
```diff
@@ -898,10 +901,9 @@ private void ExtractGeometry(BabylonAbstractMesh babylonAbstractMesh, List<Globa
                         }
                     }
                 }
-
+
```
Drigax, line 904:
```suggestion

```

Drigax (approved):

3ds Max/Max2Babylon/Exporter/BabylonExporter.Mesh.cs
```diff
@@ -808,12 +808,15 @@ private void ExtractGeometry(BabylonAbstractMesh babylonAbstractMesh, List<Globa
...
                 var indexCount = 0;
                 var minVertexIndex = int.MaxValue;
                 var maxVertexIndex = int.MinValue;
+                // Material Id is 0 if normal material, and GetMaterialID if multi-material
+                // default is [1,n] increment by 1 but user can decide to change the id (still an int) and set for example [4,3,9,1]
+                // note that GetMaterialID return the user id minus 1
+                int materialId = multiMatsCount == 1? i : meshNode.NodeMaterial.GetMaterialID(i);
                 var subMesh = new BabylonSubMesh { indexStart = indexStart, materialIndex = i };
-
+
```
Drigax, line 819:
```suggestion

```

Drigax (approved):

PR #930, opened 2021-02-10
Add material specific attributes

### LINKED ISSUE #923: Set material environmentIntensity in 3DS Max
Is there a way to set the material's environmentIntensity to 0 in 3DS Max before exporting, or do I need to hand-code this after loading the mesh?

pandaGaume:
Add Specific Babylon Full PBR related attributes to PhysicalMaterial. fix #923.
Add Babylon/GLTF attribute to force the use of roughness or metallic factor even when map is set. "Use Scalar Factor". Related to this [topic](https://forum.babylonjs.com/t/how-to-use-the-pbrmaterial-metallic-property-in-3ds-max-export-to-glb/18187)
Layout the attribute to get a cleaner control
image

### DIFF
--- a/3ds Max/Max2Babylon/Exporter/BabylonExporter.Material.cs
+++ b/3ds Max/Max2Babylon/Exporter/BabylonExporter.Material.cs
@@ -1148,15 +1148,27 @@ namespace Max2Babylon
                         + "\r\n" + " "
                         + "\r\n" + "rollout params \"Babylon Attributes\""
                         + "\r\n" + "("
-                        + "\r\n" + "checkbox babylonUnlit_ui \"Unlit\""
+                        + "\r\n" + "checkbox babylonUnlit_ui \"Unlit\" across:3"
                         + "\r\n" + "checkbox babylonBackfaceCulling_ui \"Backface Culling\""
-                        + "\r\n" + "spinner babylonMaxSimultaneousLights_ui \"Max Simultaneous Lights \" Align: #Left type: #integer Range:[0,100,4]"
                         + "\r\n" + "checkbox babylonUseFactors_ui \"Use Scalar Factors\""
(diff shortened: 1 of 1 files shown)


3ds Max/Max2Babylon/Exporter/BabylonExporter.Material.cs
```diff
@@ -403,14 +403,20 @@ private void ExportPhysicalMaterial(IIGameMaterial materialNode, IIPropertyConta
...
-                IIGameProperty babylonTransparencyModeGameProperty = babylonAttributesContainer.QueryProperty("babylonTransparencyMode");
-                if (babylonTransparencyModeGameProperty != null)
+                IIGameProperty gameProperty = babylonAttributesContainer.QueryProperty("babylonTransparencyMode");
+                if (gameProperty != null)
                 {
-                    babylonMaterial.transparencyMode = babylonTransparencyModeGameProperty.GetIntValue();
+                    babylonMaterial.transparencyMode = gameProperty.GetIntValue();
                     isTransparencyModeFromBabylonAttributes = true;
                 }
+                gameProperty = babylonAttributesContainer.QueryProperty("babylonUseFactors");
```
Drigax, line 415: I'm a bit confused, why do we want this flag? if a map is being used for O/R/M, don't we want to export them instead of the factors?

Drigax (changes requested):

3ds Max/Max2Babylon/Exporter/BabylonExporter.Material.cs
pandaGaume: Max behavior is Map OR scalar, but gltf and Babylon use these factor to also multiply the values of the map. According to the GLTF 2.0 Spec "If both factors and textures are present the factor value acts as a linear multiplier for the corresponding texture values."

Drigax: Makes sense, thanks for the explanation.

Drigax (approved):

PR #950, opened 2021-04-23
Optimize vs morph

### LINKED ISSUE #933: Maya optimitize vertices disabled on meshes with morph targets
When exporting a mesh with blendshapes / morph targets, and the Optimize Vertices option is enabled, the Maya exporter plugin raises a warning "Unable to optimize a mesh with morph targets. Disabling optimization for this mesh."

I'm curious what might break if this limitation is removed, maybe something to do with edge smoothing angles? I've tested removing this check and the resulting glTF appears to be valid, but it was on simple geometry.

pandaGaume:
Removing the Morph vs vertices "optimization" for Maya. Fix #933


Maya/Exporter/BabylonExporter.Mesh.cs
```diff
@@ -509,14 +509,8 @@ private BabylonNode ExportMesh(MDagPath mDagPath, BabylonScene babylonScene)
             //var hasColor = unskinnedMesh.NumberOfColorVerts > 0;
             //var hasAlpha = unskinnedMesh.GetNumberOfMapVerts(-2) > 0;
 
-            // TODO - Add custom properties
-            //var optimizeVertices = false; // meshNode.MaxNode.GetBoolProperty("babylonjsexportParameters.optimizeVertices");
+            //var optimizeVertices = meshNode.MaxNode.GetBoolProperty("babylonjsexportParameters.optimizeVertices");
             var optimizeVertices = exportParameters.optimizeVertices; // global option
-            if (optimizeVertices && hasMorphTarget)
```
Drigax, line 515: Was this fixed via your fix of the morph target rebuilding logic via https://github.com/BabylonJS/Exporters/pull/908?

Drigax (approved):

Maya/Exporter/BabylonExporter.Mesh.cs
pandaGaume: Suppose to be, the guy's use it sucessfully with a custom version removing these lines. 
However, following one of our previous meeting, we have to keep in mind that optimize vertices option HAS TO be removed because it's the normal way to do it.

PR #952, opened 2021-05-03
Null Tangent check

### LINKED ISSUE #951: 3ds Max exporter. "Object reference not set to an instance of an object" when exporting tangents
`Exception raised during export node will be exported as dummy node. Object reference not set to an instance of an object`
This started to happen on some geometry and I don't really see any common thread here - even a newly generated box with the option export tangents gets this message. Newly generated sphere doesn't. 
Anything that starts as a box primitive is 'cursed' no matter later modifiers, smoothing etc. 
Using Mesher, exporting to fbx and re-importing ... all the usually tricks to get rid of geometry junk don't work here. 
The only object I found that is also not exporting properly (apart from Box primitive) is anything that starts of as an extruded spline. 

I've gone through a lot of strange 'adventures' with 3ds Max, but this is on a completely new level. 

plugin version v20210401.5 (edit: updated to the latest version, has the same issue)

edit: I think I've solved this puzzle. Adding UnwrapUVW modifier fixed the issue.  So it needed a UV channel and the script I was using to generate boxes with the 'Generate Mapping Coords.' set to off.

pandaGaume:
When any of object tangents are null and "use tangent" check box has been set, this was result into Object Null Exception. Fix #951

### DIFF
--- a/3ds Max/Max2Babylon/Exporter/BabylonExporter.Mesh.cs
+++ b/3ds Max/Max2Babylon/Exporter/BabylonExporter.Mesh.cs
@@ -419,7 +419,10 @@ namespace Max2Babylon
                 // Export tangents if option is checked and mesh has tangents
                 if (exportParameters.exportTangents)
                 {
-                    babylonMesh.tangents = vertices.SelectMany(v => v.Tangent).ToArray();
+                    if (vertices.All(v => v.Tangent != null))
+                    {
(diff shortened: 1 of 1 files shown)


Drigax:
Something doesnt seem right here, why would we have vertex data where some of the tangents are null, but not all?

pandaGaume:
Using Max UI, this is not possible to have some null and others set, but for objects built by script, everything seems possible ...

Drigax (approved):

PR #955, opened 2021-05-07
Add 8 bone export support for Babylon->GLTF

Drigax:
Add glTF serialization of JOINTS_1 and WEIGHTS_1. Also add extra logic to prevent ACCESSOR_JOINTS_USED_ZERO_WEIGHT.


Drigax (commented):
Tested this against: [octopus_low.zip](https://github.com/BabylonJS/Exporters/files/6438879/octopus_low.zip). This looks to export without error.

PR #960, opened 2021-05-13
3dsMax : Add keep position option to multi export process.

### LINKED ISSUE #904: Using Multi-File Export insertion point is reset to the center of the figure
Hello this looks like an issue, maybe I just cant find the configuration option but for testing I am using the hotel.rvt file from the Autodeks Revit API tutorial. There I made 8 room shapes. And the idea is to export them as .gltf files separately and then when imported in other environment they will be blaced the same way as they were in 3ds Max. 
If I select each Room Shape separately and export them they stay in the same relative position to origin as they were on the blueprint. When I use the Multi-File export all 8 rooms have their insertion point reset to their center so consecutive insertion lets say in Blender results in obtaining all the rooms in the same place. 
This is how the rooms are placed in 3ds Max.
image
Exporting one by one with the option "Export only selected" and then importing in Blender:
image
Finally exporting with the Multi-File Export and then importing in blender:
image
Being able to do Multi-File Export but keeping the same reference point for further insertion would allow to export components partially and separately and being able to rebuild the same figure partially or completely. In my case I am interested to load separate gltf to a web so their positions would recreate the same scene.
Are you considering to implement that as a possible improvement? I am pretty new in this matter, at least tell me in which .cs file I should look to try and implement this functionality.

Thank you very much.

pandaGaume:
According to #904 Introduce a new option in order to be able to do Multi-File Export without reseting the position pivot to origin. Fix #904.
image

### DIFF
--- a/3ds Max/Max2Babylon/Exporter/BabylonExporter.cs
+++ b/3ds Max/Max2Babylon/Exporter/BabylonExporter.cs
@@ -476,7 +476,12 @@ namespace Max2Babylon
                 BabylonNode node = exportNodeRec(maxRootNode, babylonScene, gameScene);
                 // if we're exporting from a specific node, reset the pivot to {0,0,0}
                 if (node != null && exportNode != null && !exportNode.IsRootNode)
-                    SetNodePosition(ref node, ref babylonScene, new float[] { 0, 0, 0 });
+                {
+                    if (!exportParameters.exportKeepNodePosition)
+                    {
+                        SetNodePosition(ref node, ref babylonScene, new float[] { 0, 0, 0 });
+                    }
+                }
 
                 progression += progressionStep;
                 ReportProgressChanged((int)progression);
--- a/3ds Max/Max2Babylon/Exporter/BabylonExporter.ExportItem.cs
+++ b/3ds Max/Max2Babylon/Exporter/BabylonExporter.ExportItem.cs
@@ -102,11 +102,20 @@ namespace Max2Babylon
                 IsDirty = true;
             }
         }
+        public bool KeepPosition
+        {
+            get { return keepPosition; }
+            set
+            {
+                if (keepPosition == value) return;
+                keepPosition = value;
+                IsDirty = true;
+            }
+        }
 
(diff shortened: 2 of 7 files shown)


3ds Max/Max2Babylon/Exporter/BabylonExporter.cs
```diff
@@ -476,7 +476,12 @@ public void Export(MaxExportParameters exportParameters)
                 BabylonNode node = exportNodeRec(maxRootNode, babylonScene, gameScene);
                 // if we're exporting from a specific node, reset the pivot to {0,0,0}
                 if (node != null && exportNode != null && !exportNode.IsRootNode)
-                    SetNodePosition(ref node, ref babylonScene, new float[] { 0, 0, 0 });
+                {
+                    if (!exportParameters.exportKeepNodePosition)
```
Drigax, line 480: Why did the behavior differ between multi-export and regular export?

pandaGaume: This is THE question.. I do not know. It's not obvious to see the meaning behind these lines. Additionaly, another question is why there is two differents path (differents properties into the exporter ) to export single or multi-object.. Here, my proposal is only about a fix to give user a better experience without excessive changes. 
In term of software, i'm agree, it's ugly 🥇 .

Drigax: hmm, I think it may be best to understand why the behavior differs, instead of making a spot fix here. If we're going down separate codepaths for the different export methods, we should probably figure out how to combine those paths, I was under the impression that multiexport was merely a batching of export commands, and shouldnt change behavior...

pandaGaume: I already conduct this research and did not find any informations into the code. It's like this is a quick and dirty response to a feature request... The only noticiable diff is about the export parameters augmented with a node property, which lead to trigger these lines.
We may have to search down the PR history to see when it has been introduced.

Drigax (approved):

PR #968, opened 2021-06-10
Fixes event issue on adding attribute to normalMaps

pandaGaume:
In some cases, when loading a scene, the new automatic way of adding custom attributes to Normal Bump leads to a stack overflow (mainly due to multi-threaded racing). In addition, this method was too cumbersome when a scene had a lot of material, degrading performance.
We modify this behavior by adding a specific button to the materials panel.
image

### DIFF
--- a/3ds Max/Max2Babylon/MaterialScripts.cs
+++ b/3ds Max/Max2Babylon/MaterialScripts.cs
@@ -20,41 +20,28 @@ namespace Max2Babylon
           l = custAttributes.count mat; 
           for i = 1 to l+1 do ( 
             if(mat.custAttributes[i] != undefined) then (
-                if(mat.custAttributes[i].name == attName) then return true
-            ) ) return false )";
+                if(mat.custAttributes[i].name == attName) then
+                ( 
+                    return true
+                )
+            ) )  return false )";
 
-        /// <summary>
-        /// Note : We may run a check on all Slate view, because race condition into script execution 
-        /// do not ensure the link beetwen material and bump are correctly set when material is updated.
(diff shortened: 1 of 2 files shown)


3ds Max/Max2Babylon/MaterialScripts.cs
```diff
@@ -20,41 +20,28 @@ public static class MaterialScripts
...
+        		))))";
 
-        /// There is NO WAY to get an even on BumpMap creation into the Slate editor. Only Material creation is triggered.
-        /// This is the reason we listen for parameters Material change
         private static string AddPhysicalBabylonUI =
         $@"if ( hasBabylonCustomAttribute maxMaterial ""Babylon Attributes"" == false) then ( 
-              {PhysicalBabylonCAtDef}
+             {PhysicalBabylonCAtDef}
               custAttributes.add maxMaterial PHYSICAL_MATERIAL_CAT_DEF 
-              when parameters maxMaterial change do checkBumpMap()
```
Drigax, line 57: I'm assuming that this is what's causing our stackoverflow? We change the material when we execute our script checkBumpMap()? I would hope that there's some other workaround than making this behavior explicit... i.e this makes a workflow where one is programmatically creating materials a bit more cumbersome, as now one has to make an additional step to add the material custom attributes as well...

If it can't be helped, then hey, its a recurring theme that specifying material attributes outside the capabilities of the DCC's material model is not fun, regardless of the DCC tool it seems...

Whatever solution you choose to proceed with should work

Drigax (approved):

PR #969, opened 2021-06-10
update epsilon to smaller value

### LINKED ISSUE #964: Lowered Epsilon value
It's common into scene design to set a scene unit as milimeters. In this case, the Epsilon threshold which is use by Tools, is a way too small when manipulated for gltf math purpose (meters). This is affect especially Morphs and animation . For examples, when optimizing animation, it's drive to loose some significant animation.
I propose for make a simple fix and set a value of 1/1000000 instead of 1/1000.

`        public const float Epsilon = 0.001f;
`
to become
`        public const float Epsilon = 0.000001f;
`

pandaGaume:
fix #964.

### DIFF
--- a/3ds Max/Max2Babylon/Tools/Tools.cs
+++ b/3ds Max/Max2Babylon/Tools/Tools.cs
@@ -225,7 +225,7 @@ namespace Max2Babylon
             return (IIGameMorpher)constructor.Invoke(new object[] { obj.GetNativeHandle(), false });
         }
 
-        public const float Epsilon = 0.001f;
+        public const float Epsilon = 0.000001f;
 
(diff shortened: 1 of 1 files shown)


Drigax (approved):
This seems like a kicking of the can, I know that we use Tools.Epsilon in different places in the codebase, however maybe we're overloading this, as we may want different tolerances between animation diffs, vertex attributes, etc..

That said, 0.001 seems a bit high for an epsilon, so i agree that reducing it makes sense in this respect, but similar to @bghgary's concerns, we should probably get a better idea of everything that depends on this epsilon so we can know whether or not we may regress behavior elsewhere by reducing it.

I'll approve to not block merging, but we should take some time to understand whether or not there was a reason why we aren't just using Float.Epsilon...

PR #987, opened 2021-07-20
correct KHR_TextureTransform warn

### LINKED ISSUE #986: Max2Babylon: why warn for disabled KHR_texture_transform if no transforms?
Is there a reason why the exporter warns about a disabled KHR_texture_transform when the bitmaps have no transforms applied (default behavior for bitmaps)?

image

3ds Max 2020.3
Max2Babylon v20210624.1

Thanks!

pandaGaume:
the exporter warns about a disabled KHR_texture_transform when the bitmaps have no transforms applied.
VScale is set to -1 for Babylon export then the sign is flipped again for the Babylon2Gltf transformation.
However prior to this, the test for the KHR_TextureTransform was made against 1 for Babylon.VScale.
Also, we remove the duplicate warning message.
fix #986

### DIFF
--- a/SharedProjects/Babylon2GLTF/GLTFExporter.Texture.cs
+++ b/SharedProjects/Babylon2GLTF/GLTFExporter.Texture.cs
@@ -259,7 +259,7 @@ namespace Babylon2GLTF
                     gltfTextureInfo.scale = babylonTexture.level;
                 }
 
-                if (!(babylonTexture.uOffset == 0) || !(babylonTexture.vOffset == 0) || !(babylonTexture.uScale == 1) || !(babylonTexture.vScale == 1) || !(babylonTexture.wAng == 0))
+                if (!(babylonTexture.uOffset == 0) || !(babylonTexture.vOffset == 0) || !(babylonTexture.uScale == 1) || !(babylonTexture.vScale == -1) || !(babylonTexture.wAng == 0))
                 {
                     // Add texture extension if enabled in the export settings
                     if (exportParameters.enableKHRTextureTransform)
@@ -268,7 +268,6 @@ namespace Babylon2GLTF
                     }
                     else
                     {
-                        logger.RaiseWarning("GLTFExporter.Texture | KHR_texture_transform is not enabled, so the texture may look incorrect at runtime!", 3);
                         logger.RaiseWarning("GLTFExporter.Texture | KHR_texture_transform is not enabled, so the texture may look incorrect at runtime!", 3);
                     }
                 }
--- a/SharedProjects/Babylon2GLTF/GLTFExporter.Texture.cs
+++ b/SharedProjects/Babylon2GLTF/GLTFExporter.Texture.cs
@@ -259,18 +259,8 @@ namespace Babylon2GLTF
                     gltfTextureInfo.scale = babylonTexture.level;
                 }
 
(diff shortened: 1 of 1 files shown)


SharedProjects/Babylon2GLTF/GLTFExporter.Texture.cs
```diff
@@ -259,7 +259,7 @@ private GLTFTextureInfo ExportTexture(BabylonTexture babylonTexture, GLTF gltf,
                     gltfTextureInfo.scale = babylonTexture.level;
                 }
 
-                if (!(babylonTexture.uOffset == 0) || !(babylonTexture.vOffset == 0) || !(babylonTexture.uScale == 1) || !(babylonTexture.vScale == 1) || !(babylonTexture.wAng == 0))
+                if (!(babylonTexture.uOffset == 0) || !(babylonTexture.vOffset == 0) || !(babylonTexture.uScale == 1) || !(babylonTexture.vScale == -1) || !(babylonTexture.wAng == 0))
```
bghgary, line 262: Maybe it makes more sense to put this check inside the AddTextureTransformExtension after it inverts the y axis back?

bghgary (approved):

SharedProjects/Babylon2GLTF/GLTFExporter.Texture.cs
pandaGaume: Yes, but in this case, as it has been designed, we loose the opportunity to warn the user when KHR_TextureTransform is unchecked. 
However, i do not understand why we have still this KHR_TextureTransform checkBox. This maybe has sense long time ago when the existing viewers did not support this extension.. but today, it's a non sense...

bghgary: You can just pass logger into the function.

bghgary: Wait, I thought logger is a variable, but maybe it's a namespace. Why do you loose the opportunity to warn the user?

pandaGaume: logger is a field of the GLTFExporter partial class, so it's available into the method
`private void AddTextureTransformExtension(..)`
Without logic change, this method is called only if use KHR_TextureTransform checkbox is checked.
The warning is raised only when unchecked.
```
if( Has Transform )
{
  if( enableKHRTextureTransform checked)
  {
     AddTextureTransformExtension()
  } 
 else 
 {
     warn the user
  }
}
```

We can for sure move the code, but this will change the actual logic and then we might want to change the method name to reflect the fact we MAY add the extension. 

I do not wrote this code, so either way are fine for me, i just choose the one with the less changes.

bghgary: This was just a comment. I'm okay if you merge it as is. It would be better if we didn't have the invert logic in multiple places.

pandaGaume: You're right, it's better to group logic. Here the new proposal.

bghgary: looks good

PR #1039, opened 2022-03-18
Add some null checks to make it work with Blender-exported meshes and…

carolhmj:
… make it more resilient.

### DIFF
--- a/Tools/MakeIncremental/Program.cs
+++ b/Tools/MakeIncremental/Program.cs
@@ -211,7 +211,7 @@ namespace MakeIncremental
                 }
 
                 dynamic scene;
-                var outputDir = Path.GetDirectoryName(dstPath);
+                var outputDir = (dstPath != null && dstPath != "") ? Path.GetDirectoryName(dstPath) : Path.GetDirectoryName(input);
                 var rootFilename = Path.GetFileNameWithoutExtension(input);
 
                 // Loading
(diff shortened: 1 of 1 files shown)


Tools/MakeIncremental/Program.cs
```diff
@@ -211,7 +211,7 @@ static void ProcessSourceFile(string input, string dstPath)
                 }
 
                 dynamic scene;
-                var outputDir = Path.GetDirectoryName(dstPath);
+                var outputDir = (dstPath != null && dstPath != "") ? Path.GetDirectoryName(dstPath) : Path.GetDirectoryName(input);
```
bghgary, line 214:
```suggestion
                var outputDir = String.IsNullOrEmpty(dstPath) ? Path.GetDirectoryName(input) : Path.GetDirectoryName(dstPath);
```

bghgary (commented):
I don't know this code that well. How do you test these changes?

Tools/MakeIncremental/Program.cs
pandaGaume: @carolhmj could we resolve ?

carolhmj:
@pandaGaume updated now, sorry I missed that!

carolhmj:
@bghgary about the testing, I found the problem while trying to use it on the basic cube exported from Blender: https://forum.babylonjs.com/t/incremental-babylon-guide-for-dummies/28453/21?u=carolhmj

bghgary:
> @bghgary about the testing, I found the problem while trying to use it on the basic cube exported from Blender: https://forum.babylonjs.com/t/incremental-babylon-guide-for-dummies/28453/21?u=carolhmj

Okay. I'm assuming both code paths of the `outputDir` was tested. I don't know this code well enough to determine if the code is correct.

bghgary:
@carolhmj Do we still want this?

carolhmj:
@bghgary hmmm I haven't seen anyone complaining about this so I guess it's not needed? We can reopen if needed.

bghgary:
Okay, closing for now. Please reopen if needed.

PR #1056, opened 2022-06-17
Remove mandatory Maya rename operation 

pandaGaume:
Remove mandatory rename operation for Maya plug-in, by updating the build scripts.


bghgary:
I disabled the rename in the build pipeline. Try again.

bghgary:
The one I manually kicked off succeeded, but the CI won't use the new pipeline definition until something changes in the PR.

Issue #1042, opened 2022-04-01
Maya Export Texture Formats


TyLindberg:
Would it be possible to add the functionality from #887 to the Maya exporter as well.

Right now there's no way to export a png ORM texture that I can identify.

pandaGaume:
I will have to get back on Maya very soon to finalize the support of the GLTF Extensions, then i will be able to dig on this topic.
G.

PR #1074, opened 2022-09-09
Expose texture format policy as comboBox for Max and Maya ( issue #1042)

BenIngoldVV:
Exposing textureFormatPolicy (from https://github.com/BabylonJS/Exporters/pull/887) in form as a combobox
Done also for Maya (resolving https://github.com/BabylonJS/Exporters/issues/1042)

### DIFF
--- a/3ds Max/Max2Babylon/Exporter/MaxExportParameters.cs
+++ b/3ds Max/Max2Babylon/Exporter/MaxExportParameters.cs
@@ -29,7 +29,6 @@ namespace Max2Babylon
         public bool mergeContainersAndXRef = false;
         public bool flattenScene = false;
         public BakeAnimationType bakeAnimationType = BakeAnimationType.DoNotBakeAnimation;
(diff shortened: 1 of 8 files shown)


pandaGaume (changes requested):
Hey, many thanks for the PR.
The initial design of the exporter was done using Windows Form instead at WPF, with raw location of component  instead of layout, Consequently, the change MUST be made into screen resolution of 1024x768 to keep the form usable on every definition (VSStudio and design windows are working with the native resolution of the screen). I know this is "a bit" anoying and this is the reason i do not touched the forms so far and ADD an advanced  panel which is using layout. 
You can use this advanced panel or change your screen resolution prior of the form modification.

BenIngoldVV:
Hey,
Thanks for the review, I've saved again with a 1024*768 screen res. I don't know how to check if I've done this right

pandaGaume (approved):
sounds good

pandaGaume (changes requested):
Avoid Null pointer exception with Draco options

bghgary:
@BenIngoldVV Do you still want this PR to go in?

PR #1113, opened 2024-03-12
Add support for 8 uv channels

SergioRZMasson:
Increased the number of supported UV channels from 2 to 8.


Maya/Exporter/BabylonExporter.Mesh.cs
```diff
@@ -939,6 +963,102 @@ private GlobalVertex ExtractVertex(MFnMesh mFnMesh, int polygonId, int vertexInd
...
+                    float u = 0, v = 0;
+                    mFnMesh.getPolygonUV(polygonId, vertexIndexLocal, ref u, ref v, uvSetNames[indexUVSet]);
+                    vertex.UV3 = new float[] { u, v };
+                }
+                catch
+                {
+                    // An exception is raised when a vertex isn't mapped to an UV coordinate
+                    isUVExportSuccess[indexUVSet] = false;
+                }
+            }
```
bghgary, line 980: Use a helper function?

bghgary: Generally, all of these repeat codes can use some helpful functions. Not a big deal, but it's a lot of duplication.

Maya/Exporter/BabylonExporter.Mesh.cs
```diff
@@ -15,7 +15,7 @@ internal partial class BabylonExporter
         private MStringArray allMayaInfluenceNames;     // the joint names that influence the mesh (joint with 0 weight included)
         private MDoubleArray allMayaInfluenceWeights;   // the joint weights for the vertex (0 weight included)
         private Dictionary<string, int> indexByNodeName = new Dictionary<string, int>();    // contains the node (joint and parents of the current skin) fullPathName and its index
-
+        private const int MAX_UV_EXPORTS = 8; // Maximum number of UVs to export.
```
bghgary, line 18: Is this useful? If we change this number, the code still won't work right?

bghgary (approved):

PR #1118, opened 2024-08-31
Installer support for 3ds max 2024/2025

FrogsInSpace:
add Babylonjs Installer support for 3ds Max 2024 and 3ds Max 2025


SergioRZMasson:
I'm not sure adding 2025 would make sense since we don't currently provide a plugin for that.

FrogsInSpace:
@SergioRZMasson i agree. 
While updating the plugincode for 3ds Max 2025, i ran over a major roadblock due to removed functionality in the 3ds Max NetSDK dealing with actiontables and menuhandling.
After that reoadblock is removed it makes sense then...

SergioRZMasson:
Sure, but I think we can keep the changes that add support for Max 2024, since we do have support for that.

FrogsInSpace:
I solved the roadblock regarding 3ds Max 2025's new menumanager API and created a new PR
https://github.com/BabylonJS/Exporters/pull/1121
the new one also includes the installer update so this one can be closed

PR #1125, opened 2024-10-30
Removed unecessary copy of 3DS Max binary

SergioRZMasson:
Setting flag so that 3DS Max binary does not get installed


bghgary (approved):

FrogsInSpace:
I created a PR (#1126)  which also includes this fix and ( more important) fixes the wrong UTF8 encoding which breaks Menu creation in 3ds Max 2025. 
Also contains several other fixes for longstanding issues

sebavan (approved):

PR #1129, opened 2024-12-19
3ds Max: 8 UV Channels

JBoye:
Increased number of UV Channels, to match the Maya Exporter


bghgary (approved):
LGTM. @SergioRZMasson Can you verify?

SergioRZMasson (approved):
Yep, that should do it.

PR #1132, opened 2025-01-27
Add check for invalid tangent data in Maya 

### LINKED ISSUE #1131: Maya 2023 Exporter can produce glb files with validation errors.
A user on the forum has provided a file that, when exported using Maya 2023 exporter plugin, generates a glb with validation errors. 

Forum  [issue](https://forum.babylonjs.com/t/babylon-js-cannot-export-due-to-newtonsoft-json/56075/6)

SergioRZMasson:
It is possible that Maya returns a vector with values 0,0,0 when exporting tangent data for mesh.  If mesh has invalid model data we should not export it to avoid creating an invalid GLB. 

Fix #1131

### DIFF
--- a/Maya/Exporter/BabylonExporter.Mesh.cs
+++ b/Maya/Exporter/BabylonExporter.Mesh.cs
@@ -919,15 +919,24 @@ namespace Maya2Babylon
                 {
                     MVector tangent = new MVector();
                     mFnMesh.getFaceVertexTangent(polygonId, vertexIndexGlobal, tangent);
+                    tangent.normalize();
+                    
+                    if (tangent.length < 0.1)
+                    {
+                        isTangentExportSuccess = false;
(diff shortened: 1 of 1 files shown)


Maya/Exporter/BabylonExporter.Mesh.cs
```diff
@@ -919,15 +919,24 @@ private GlobalVertex ExtractVertex(MFnMesh mFnMesh, int polygonId, int vertexInd
                 {
                     MVector tangent = new MVector();
                     mFnMesh.getFaceVertexTangent(polygonId, vertexIndexGlobal, tangent);
+                    tangent.normalize();
+                    
+                    if (tangent.length < 0.1)
```
bghgary, line 924: How can length be less than 0.1 after normalization? Should you check this before calling `normalize`?

PR #1137, opened 2025-04-22
Added support for 16bit bone indexes  

SergioRZMasson:
GLTF allows bone indexes to be stored as 16 bit values, in our implementation we use a 32 bit integer to store 4 bones as bytes, this breaks the exporter for models that have more than 256 bones  even though they would be able to fit into a GLTF. I'm increasing our internal store of the bones to be done using 64bit integer to allow bones to be hold using 16bit, matching what is expected by GLTF.

### DIFF
--- a/3ds Max/Max2Babylon/Exporter/BabylonExporter.cs
+++ b/3ds Max/Max2Babylon/Exporter/BabylonExporter.cs
@@ -767,6 +767,11 @@ namespace Max2Babylon
             babylonScene.Prepare(false, false);
             if (isBabylonExported)
             {
+                if (!babylonScene.PackIndexArrays())
+                {
+                    RaiseWarning("Model has too many skeleton joints. The .babylon file will store joint indicies using expandanded form, this will only load properly using Babylon.js >= 8.5.0.");
+                }
+                
                 RaiseMessage("Saving to output file");
 
                 var outputFile = Path.Combine(outputBabylonDirectory, outputFileName);
--- a/3ds Max/Max2Babylon/Exporter/BabylonExporter.cs
+++ b/3ds Max/Max2Babylon/Exporter/BabylonExporter.cs
@@ -767,7 +767,7 @@ namespace Max2Babylon
             babylonScene.Prepare(false, false);
             if (isBabylonExported)
             {
-                if (!babylonScene.PackIndexArrays())
+                if (!babylonScene.TryPackIndexArrays())
                 {
                     RaiseWarning("Model has too many skeleton joints. The .babylon file will store joint indicies using expandanded form, this will only load properly using Babylon.js >= 8.5.0.");
                 }
--- a/3ds Max/Max2Babylon/Exporter/GlobalVertex.cs
+++ b/3ds Max/Max2Babylon/Exporter/GlobalVertex.cs
@@ -19,9 +19,9 @@ namespace Max2Babylon
         public IPoint2 UV6 { get; set; }
         public IPoint2 UV7 { get; set; }
         public IPoint2 UV8 { get; set; }
-        public int BonesIndices { get; set; }
+        public long BonesIndices { get; set; }
         public IPoint4 Weights { get; set; }
-        public int BonesIndicesExtra { get; set; }
+        public long BonesIndicesExtra { get; set; }
         public IPoint4 WeightsExtra { get; set; }
         public float[] Color { get; set; }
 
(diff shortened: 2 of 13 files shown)


3ds Max/Max2Babylon/Exporter/BabylonExporter.cs
```diff
@@ -767,6 +767,11 @@ public void Export(MaxExportParameters exportParameters)
             babylonScene.Prepare(false, false);
             if (isBabylonExported)
             {
+                if (!babylonScene.PackIndexArrays())
+                {
+                    RaiseWarning("Model has too many skeleton joints. The .babylon file will store joint indicies using expandanded form, this will only load properly using Babylon.js >= 8.5.0.");
```
bghgary, line 772:
```suggestion
                    RaiseWarning("Model has too many skeleton joints. The .babylon file will store joint indices using expanded form, this will only load properly using Babylon.js >= 8.5.0.");
```

SharedProjects/BabylonExport.Entities/BabylonScene.cs
```diff
@@ -218,5 +218,17 @@ public void Prepare(bool generateDefaultLight = true, bool generateDefaultCamera
                 animationGroups = null;
             }
         }
+
+        internal bool PackIndexArrays()
```
bghgary, line 222: nit: Naming is inconsistent with the one in mesh. Maybe both should start with `Try` or maybe don't use this prefix because this code isn't actually following the `Try` prefix pattern?

Maya/Exporter/BabylonExporter.Mesh.cs
```diff
@@ -1018,15 +1019,15 @@ private GlobalVertex ExtractVertex(MFnMesh mFnMesh, int polygonId, int vertexInd
                 // decreasing sort
                 OrderByDescending(ref weightByInfluenceIndex);
 
-                int bonesCount = indexByNodeName.Count; // number total of bones/influences for the mesh
+                long bonesCount = indexByNodeName.Count; // number total of bones/influences for the mesh
```
bghgary, line 1022: Are these int --> long changes still needed?

3ds Max/Max2Babylon/Exporter/GlobalVertex.cs
```diff
@@ -19,9 +19,9 @@ public struct GlobalVertex
         public IPoint2 UV6 { get; set; }
         public IPoint2 UV7 { get; set; }
         public IPoint2 UV8 { get; set; }
-        public int BonesIndices { get; set; }
+        public int[] BonesIndices { get; set; }
```
bghgary, line 22: Maybe these should be `ushort` to begin with? It can't be bigger than this type right?

Maya/Exporter/GlobalVertex.cs
```diff
@@ -17,9 +18,9 @@ public struct GlobalVertex
         public float[] UV6 { get; set; } // Vec2
         public float[] UV7 { get; set; } // Vec2
         public float[] UV8 { get; set; } // Vec2
-        public int BonesIndices { get; set; }
+        public int[] BonesIndices { get; set; }
```
bghgary, line 21: ditto

SharedProjects/Babylon2GLTF/GLTFGlobalVertex.cs
```diff
@@ -16,8 +16,8 @@ public class GLTFGlobalVertex
         public BabylonVector2 UV7 { get; set; }
         public BabylonVector2 UV8 { get; set; }
         public float[] Color { get; set; }
-        public ushort[] BonesIndices { get; set; }
-        public ushort[] BonesIndicesExtra { get; set; }
+        public int[] BonesIndices { get; set; }
+        public int[] BonesIndicesExtra { get; set; }
         public float[] BonesWeights { get; set; }
```
bghgary, line 21: Why change this?

SharedProjects/BabylonExport.Entities/BabylonScene.cs
```diff
@@ -218,5 +218,17 @@ public void Prepare(bool generateDefaultLight = true, bool generateDefaultCamera
...
+        {
+            bool result = true;
+
+            foreach (var mesh  in meshes)
```
bghgary, line 226: nit: extra space
```suggestion
            foreach (var mesh in meshes)
```

bghgary (approved):

PR #1141, opened 2025-07-23
Add ability to set useAlphaFromDiffuseTexture from UI

vladimir-tikhonov:
image

Pretty much self-explanatory. Another approach would be to set it conditionally, like useAlphaFromAlbedoTexture, but I don’t feel like introducing a breaking change into a project that's in maintenance mode :) It doesn’t take up much space in the UI and might be useful for someone - in fact, it’s already beneficial for our team


SergioRZMasson (approved):
Looks good! Thanks for the contribution!

